PR #44, opened 2020-03-27
Implement updates to reflect latest reference data schema and add local entity name support

sgardnerdell-ea:


graham-willis-druid:
Looks good

PR #75, opened 2020-04-08
Add contact details pages to the web journey 

graham-willis-druid:


packages/gafl-webapp-service/src/lib/address-lookup-service.js
```diff
@@ -0,0 +1,47 @@
...
+    dataset: ADDRESS_LOOKUP_SERVICE.dataset
+  })
+
+  url.search = params.toString()
+
+  debug({ url })
+
+  const { results } = await (async () => {
+    try {
+      const response = await fetch(url.href, {
```
sgardnerdell-ea, line 23: Should be setting a timeout here to prevent the incoming request from timing out if the lookup service fails to respond in time.

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -21,8 +21,15 @@ export default (path, view, completion, getData) => ({
     const page = await request.cache().helpers.page.getCurrentPermission(view)
     const pageData = page || {}
     if (getData && typeof getData === 'function') {
-      const data = await getData(request)
-      Object.assign(pageData, { data })
+      try {
+        const data = await getData(request)
+        Object.assign(pageData, { data })
+      } catch (err) {
```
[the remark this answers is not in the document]
graham-willis-druid: Yes its a todo - I was going to tackle it when I did the coverage. I will probably log the error - it needs a mock

packages/gafl-webapp-service/src/lib/address-lookup-service.js
graham-willis-druid: Yes actually good point - I think a 10 second timeout would be appropriate but I will add it as an environment variable

graham-willis-druid:
Closing because the sands are shifting. I am going to integrate the business rules stuff

PR #89, opened 2020-04-16
Add summary page: personal details functionality

graham-willis-druid:


packages/gafl-webapp-service/src/processors/__tests__/concession-helper-spec.js
```diff
@@ -0,0 +1,89 @@
...
+    f.addDisabled(licensee, CONCESSION_PROOF.blueBadge, '123')
+    expect(licensee.concessions).toContainEqual(disabledbb)
+    expect(licensee.concessions.length).toBe(1)
+  })
+  it('add NI', () => {
+    f.addDisabled(licensee, CONCESSION_PROOF.NI, '456')
+    expect(licensee.concessions).toContainEqual(disabledNi)
+    expect(licensee.concessions.length).toBe(1)
+  })
+  it('add junior', () => {
```
sgardnerdell-ea, line 57: There are a number of duplicate/repeated tests here

packages/gafl-webapp-service/src/processors/concession-helper.js
```diff
@@ -0,0 +1,105 @@
+import { CONCESSION, CONCESSION_PROOF } from './mapping-constants.js'
```
sgardnerdell-ea, line 1: It looks like each of the addXXX methods here always remove the other type of concessions?  If so, would it not be simpler to just recreate the array each time?

packages/gafl-webapp-service/src/processors/concession-helper.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    })
+  }
+}
+
+const hasJunior = licensee => licensee.concessions && licensee.concessions.find(c => c.type === CONCESSION.JUNIOR)
+
+const removeJunior = licensee => {
+  if (hasJunior(licensee)) {
+    licensee.concessions.splice(
```
sgardnerdell-ea, line 34: More efficient to do `licensee.concessions = licencee.concessions.filter(c => c.type !== CONCESSION_JUNIOR)`

packages/gafl-webapp-service/src/processors/concession-helper.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+    })
+  }
+}
+
+const hasSenior = licensee => licensee.concessions && licensee.concessions.find(c => c.type === CONCESSION.SENIOR)
+
+const removeSenior = licensee => {
+  if (hasSenior(licensee)) {
+    licensee.concessions.splice(
```
sgardnerdell-ea, line 65: As per comment on removeJunior

packages/gafl-webapp-service/src/processors/concession-helper.js
```diff
@@ -0,0 +1,105 @@
...
+      type: concessionProof,
+      referenceNumber: referenceNumber
+    }
+  })
+}
+
+const hasDisabled = licensee => licensee.concessions && licensee.concessions.find(c => c.type === CONCESSION.DISABLED)
+
+const removeDisabled = licensee => {
+  if (hasDisabled(licensee)) {
```
sgardnerdell-ea, line 97: As per comment on removeJunior

packages/gafl-webapp-service/src/processors/get-permit.js
```diff
@@ -0,0 +1,34 @@
...
+  // Filter the joined list to include every and only those concessions in licenseeConcessions
+  const filteredPermitsJoinPermitConcessions = permitsJoinPermitConcessions.filter(
+    pjpc =>
+      licenseeConcessions.map(lc => lc.type).every(t => pjpc.concessions.map(c => c.name).includes(t)) &&
+      pjpc.concessions.length === licenseeConcessions.length
+  )
+
+  // Filter by the licence length
+  const byLicenceLength = filteredPermitsJoinPermitConcessions.filter(
+    p => String(p.durationMagnitude + p.durationDesignator.label) === permission.licenceLength
```
sgardnerdell-ea, line 24: FYI - as per the developer catchup meeting this will be changing to `p.durationDesignator.description` shortly

packages/gafl-webapp-service/src/server.js
```diff
@@ -75,7 +75,6 @@ const init = async () => {
   const sessionCookieName = process.env.SESSION_COOKIE_NAME || SESSION_COOKIE_NAME_DEFAULT
 
   const sessionCookieOptions = {
```
sgardnerdell-ea, line 77: When will the session cookie expire?  It needs to have a ttl to align with the new PECR cookie guidance rules.

packages/gafl-webapp-service/src/services/sales-api/sales-api-service.js
```diff
@@ -0,0 +1,35 @@
...
+    })
+    return response.json()
+  } catch (err) {
+    // On a failure to connect do not stop the user journey
+    console.error('Unable to connect to the sales API service', err)
+    throw err
+  }
+}
+
+const permitsOperations = {
```
sgardnerdell-ea, line 20: These URL's changed recently, remove the `/reference-data` segment from the path.

Also could be simplified to:

```
  fetchPermits: async() => fetchData(new URL('/permits', urlBase)),
  fetchConcessions: async() => fetchData(new URL('/concessions', urlBase)),
  fetchPermitConcessions: async() => fetchData(new URL('/permitConcessions', urlBase))
```

packages/gafl-webapp-service/src/processors/__tests__/concession-helper-spec.js
graham-willis-druid: yes - the repeated tests are the combinations needed to cover all the lines

packages/gafl-webapp-service/src/processors/concession-helper.js
graham-willis-druid: It would be I agree but the thought was that this should basically cover the situation where there are multiple concessions without to much change

packages/gafl-webapp-service/src/processors/concession-helper.js
graham-willis-druid: ok I will fix

packages/gafl-webapp-service/src/processors/concession-helper.js
graham-willis-druid: ditto

packages/gafl-webapp-service/src/services/sales-api/sales-api-service.js
graham-willis-druid: I know

graham-willis-druid: Will fix

packages/gafl-webapp-service/src/server.js
graham-willis-druid: At the moment has an expiry  of session. I do not know if this is compliant or not but basically the cache is doing the timing out. Its tricky otherwise because the cache ttl gets updated on every write whereas the cookie just runs down. I need to return to this I guess

PR #136, opened 2020-05-11
Integrate GOV.UK Pay into the web journey

graham-willis-druid:


packages/gafl-webapp-service/src/pages/licence-details/licence-start-date/route.js
```diff
@@ -1,4 +1,6 @@
-import { LICENCE_START_DATE, CONTROLLER } from '../../../constants.js'
+import { LICENCE_START_DATE, CONTROLLER } from '../../../uri.js'
+import { ADVANCED_PURCHASE_MAX_DAYS } from '@defra-fish/business-rules-lib/src/constants.js'
```
sgardnerdell-ea, line 2: This is importing from an explicit src file rather than the globally exported index file - see my other comment about exports in the business-rules-lib/index.js file

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
```diff
@@ -50,14 +50,22 @@ const getData = async request => {
 
   await findPermit(permission, request)
 
-  // Is this date format ok?
-  const startTimeString = moment(permission.licenceStartDate, 'YYYY-MM-DD')
-    .add(permission.licenceStartTime, 'hours')
-    .format('LLLL')
+  const startDateString = moment(permission.licenceStartDate, 'YYYY-MM-DD').format('dddd, MMMM Do, YYYY')
```
sgardnerdell-ea, line 53: Shouldn't be using Midnight and Midday - I appreciate this is a little up in the air at the moment but the improvements made to the current service to use 12:01 etc were well received in user feedback.
We should use this as a starting point until there is evidence for something better.

packages/gafl-webapp-service/src/routes/__tests__/add-permission-route.spec.js
```diff
@@ -1,5 +1,6 @@
 import { start, stop, initialize, injectWithCookie } from '../../__mocks__/test-utils.js'
-import { ADD_PERMISSION, NEW_TRANSACTION, MAX_PERMISSIONS, TEST_TRANSACTION } from '../../constants.js'
+import { ADD_PERMISSION, NEW_TRANSACTION, TEST_TRANSACTION } from '../../uri.js'
+import { MAX_PERMISSIONS_PER_TRANSACTION } from '@defra-fish/business-rules-lib/src/constants.js'
```
sgardnerdell-ea, line 3: See other comments about imports/exports of constants from business-rules-lib

packages/gafl-webapp-service/src/session-cache/add-permission.js
```diff
@@ -1,6 +1,7 @@
-import { MAX_PERMISSIONS } from '../constants.js'
+import { MAX_PERMISSIONS_PER_TRANSACTION } from '@defra-fish/business-rules-lib/src/constants.js'
```
sgardnerdell-ea, line 1: See other comments about imports/exports of constants from business-rules-lib

packages/gafl-webapp-service/src/uri.js
```diff
@@ -0,0 +1,84 @@
...
+const CLIENT_ERROR = { uri: '/buy/client-error', page: 'client-error' }
+const SERVER_ERROR = { uri: '/buy/server-error', page: 'server-error' }
+
+/**
+ * These are inserted at runtime by the test framework but the session manager needs to know about them
+ */
+const TEST_STATUS = { uri: '/buy/status' }
+const TEST_TRANSACTION = { uri: '/buy/transaction' }
+
+export {
```
sgardnerdell-ea, line 50: Use `export const` on above lines instead of exporting here for brevity

packages/sales-api-service/src/schema/transaction.schema.js
```diff
@@ -2,12 +2,14 @@ import Joi from '@hapi/joi'
 import { createPermissionSchema, createPermissionResponseSchema } from './permission.schema.js'
 import { contactSchema } from './contact.schema.js'
 import { createOptionSetValidator } from './validators/index.js'
+import { MAX_PERMISSIONS_PER_TRANSACTION } from '@defra-fish/business-rules-lib/src/constants.js'
```
sgardnerdell-ea, line 5: As per other comments on imports/exports from business-rules-lib

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
graham-willis-druid: Its wrong though isn't it. The licence starts at 12:00. That is when you have asked for it to start.

graham-willis-druid: You have asked for it to start my selecting the 'Midnight' radio button. Lets pass it to Jon

graham-willis-druid: I am uncomfortable doing this. Regardless of any 'user research' or previously established rule this is incorrect a priori. On a previous page you have given the user the option to select midday as a licence start time and this is what should be played back. IMHO it looks ridiculous to then translate that into 12:01pm. If it is 12:01 pm then the radio button should be label that way. Let us tell the user accurately when his licence starts and ends please.

packages/gafl-webapp-service/package.json
```diff
@@ -36,7 +36,7 @@
     "prepare": "gulp --gulpfile build/gulpfile.cjs"
   },
   "dependencies": {
-    "@defra-fish/business-rules-lib": "^0.0.1-beta.28",
+    "@defra-fish/business-rules-lib": "0.0.1-beta.28",
```
sgardnerdell-ea, line 39: Shouldn't modify these as they are managed by lerna and modified each time a change is pushed to master.

packages/gafl-webapp-service/package.json
```diff
@@ -47,7 +47,7 @@
     "debug": "^4.1.1",
     "find": "^0.3.0",
     "govuk-frontend": "^3.6.0",
-    "moment": "^2.25.1",
+    "moment": "^2.25.3",
```
sgardnerdell-ea, line 50: To update libraries you can use the lerna update tool from the root of the repo to keep versioning consistent in all projects.
`lernaupdate --non-interactive --dependency "moment@latest"` will update all projects that use it and rewrite package.json/package-lock.json files as necessary

packages/gafl-webapp-service/src/routes/__tests__/add-permission-route.spec.js
sgardnerdell-ea: This one still needs updating to remove the `src/constants.js`

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
sgardnerdell-ea: I agree with you that it should be consistent throughout the journey and any communications but I don't think we can revert back to something that was less favourable with our users especially when we have metrics to support the change to use 12:01 etc.

sgardnerdell-ea: I've created https://github.com/DEFRA/rod-licensing/issues/137 to track this.

packages/dynamics-lib/package.json
```diff
@@ -33,13 +33,14 @@
     "test": "echo \"Error: run tests from root\" && exit 1"
   },
   "dependencies": {
+    "@defra-fish/business-rules-lib": "0.0.1-beta.28",
```
sgardnerdell-ea, line 36: This isn't required

packages/sales-api-service/package.json
```diff
@@ -35,7 +35,7 @@
     "test": "echo \"Error: run tests from root\" && exit 1"
   },
   "dependencies": {
-    "@defra-fish/business-rules-lib": "^0.0.1-beta.28",
+    "@defra-fish/business-rules-lib": "0.0.1-beta.28",
```
sgardnerdell-ea, line 38: Shouldn't be modified

packages/sqs-receiver-service/package.json
```diff
@@ -42,6 +42,7 @@
     "uuid": "^8.0.0"
   },
   "devDependencies": {
+    "@defra-fish/business-rules-lib": "0.0.1-beta.28",
```
sgardnerdell-ea, line 45: Shouldn't be modified

PR #143, opened 2020-05-20
Add package to enable post office counter licence sales to be processed

sgardnerdell-ea:
**Validators**
- Fix problem sharing validators from a common project and update all references to these in each service
- Fix titlecase problem with Lastname validator

**Connectors**
- Added shared connectors-lib project
- Added connector for AWS services and updated all existing references to AWS to use this
- Added connector for Sales API Service (currently used in POCL, frontend may be updated in future)
- Added shared mock helper for mocking AWS subsystems in each project

**GAFL Web App Service**
- Removed redundant functionality to add spacing to UK postcodes as this was already handled by the shared validators
- Updated references to shared validators

**Sales API Service**
- Added batch endpoint to process the creation of up to 25 transactions in batch
- Added support for POCL channelId
- Refactored to used shared connectors
- Updated finalise-transaction endpoint to return JSON structure
- Updated references to shared validators

**SQS Receiver Service**
- Updated to use shared connectors

**Post Office Licence Sales Processor**
- Added new POCL processor and associated staging tables
- Capability to parse a POCL XML file and transform to the payloads necessary to act on the Sales API
- Integrated with the Sales API to stage transactions
- Capability to resume if process interrupted at any point


packages/gafl-webapp-service/src/pages/licence-details/licence-start-date/__tests__/licence-start-date.spec.js
```diff
@@ -1,11 +1,4 @@
-import {
-  LICENCE_START_DATE,
-  LICENCE_LENGTH,
-  LICENCE_START_TIME,
-  DATE_OF_BIRTH,
```
graham-willis-druid, line 5: I don't think we should be making arbitrary format changes to each others code.

packages/gafl-webapp-service/src/processors/api-transaction.js
```diff
@@ -13,8 +13,9 @@ const prepareApiTransactionPayload = async request => {
     permissions: transactionCache.permissions.map(p => {
       const permission = {
         permitId: p.permit.id,
-        licensee: Object.assign((({ countryCode, ...l }) => l)(p.licensee),
-          { country: countriesList.find(c => c.code === p.licensee.countryCode).name }),
+        licensee: Object.assign((({ countryCode, ...l }) => l)(p.licensee), {
```
graham-willis-druid, line 16: I don't think we should be making arbitrary format changes to each others code.

packages/gafl-webapp-service/src/pages/contact/address/entry/route.js
```diff
@@ -5,13 +5,13 @@ import { validation } from '@defra-fish/business-rules-lib'
 import { referenceDataOperations } from '../../../../services/sales-api/sales-api-service.js'
 
 const validator = Joi.object({
-  premises: validation.contact.premisesValidator,
-  street: validation.contact.streetValidator,
-  locality: validation.contact.localityValidator,
-  town: validation.contact.townValidator,
+  premises: validation.contact.createPremisesValidator(Joi),
```
graham-willis-druid, line 8: Did we really get to the bottom of this problem? I don't think I experienced the issue in the front end, so the question still remains. I think we need to find out why. This work-around is Ok but it is rather cumbersome.

packages/pocl-job/README.md
```diff
@@ -0,0 +1,20 @@
...
+| POCL_RECORD_STAGING_TABLE | The DynamoDB table used for staging POCL records         |   yes    |           |                               |       |
+| POCL_STAGING_TTL          | The time to live for records in the either staging table |    no    | 168 hours |                               |       |
+
+###See also:
+
+- Environment variables [required by the connectors-lib package](../connectors-lib/README.md).
+
+# Prerequisites
+
+See [main project documentation](../../README.md).
```
graham-willis-druid, line 20: Can we have a description of the process. I would like to see a description the cli and an outline of the overall process. Is the transform to file supplementary?  I am presuming the JSON gets written directly into dynamo without an intermediate file

packages/pocl-job/src/staging/db.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+    await docClient.batchWrite(params).promise()
+  }
+}
+
+/**
+ * Retrieve processed records from the record staging table for the provided filename and stage filters
+ *
+ * @param {string} filename the filename of a POCL file for which the records should be retrieved
```
graham-willis-druid, line 49: filename the filename!

packages/pocl-job/src/__mocks__/test-2-records.xml
```diff
@@ -0,0 +1,122 @@
+<NewLicence xmlns="EAF1NewLicence.xsd" >
```
graham-willis-druid, line 1: Did you write a test file generator or can we reuse the one that is already in the jenkins job? What will happen when the GDPR change to the POCL file comes into being - will the loader just ignore it? Should we add it in now?

packages/gafl-webapp-service/src/pages/licence-details/licence-start-date/__tests__/licence-start-date.spec.js
sgardnerdell-ea: I think this was picked up by prettier-standard.  It's slightly odd as the whole point is that it should be consistent for us both so this sort of reformatting change should never occur.

packages/gafl-webapp-service/src/processors/api-transaction.js
sgardnerdell-ea: As above - prettier-standard picked this up

packages/pocl-job/README.md
sgardnerdell-ea: Yes, agreed.  I am doing further work on it at the moment to add the SFTP/S3 support so will include documentation at this point.

packages/pocl-job/src/__mocks__/test-2-records.xml
sgardnerdell-ea: I have not currently implemented a test file generator - these are static and are used in the unit tests.

I was in two minds whether to implement against the current schema or for the upcoming changes but we have not yet received confirmation if/when the schema changes will be made so will address that separately.

sgardnerdell-ea:
> Looks really good. I wonder how it may be extended to add the DD loader. Is the code amendable to easily to add additional loaders

I am hoping that we won't need a DD loader if we can implement a strategic solution to integrate with the SSCL API to allow new Direct Debits to be setup in addition to processing existing mandates.  I'll bring this up with Claire.

packages/pocl-job/src/staging/db.js
sgardnerdell-ea: lol - I suppose I could change to:

```
@param {string} filename the name of a POCL file for which the records should be retrieved
```

packages/gafl-webapp-service/src/pages/contact/address/entry/route.js
sgardnerdell-ea: Not entirely - it appears that Joi builds an internal list of validators and as we were previously not passing in the global Joi object it was returning the wrong validation error in the API.  I proved this by copying the shared validators into the SalesAPI project where everything immediately started working correctly but referencing via the business-rules-lib project failed.  I don't think it is worth the time to debug deep into the bowels of Joi to find the root cause.  I suspect if you were using nested/hierarchical structures in the frontend then you would've encountered this problem also.

PR #190, opened 2020-06-17
Implement Google Analytics

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-1114

Comprised of server-side analytics integration, an attribution endpoint,
enhanced ecommerce events being sent to GA  and the capability to send
page views to a second tracking service


packages/gafl-webapp-service/src/handlers/attribution-handler.js
```diff
@@ -0,0 +1,22 @@
...
+  await cache.helpers.status.set({
+    attribution: {
+      [UTM.CAMPAIGN]: request.query[UTM.CAMPAIGN],
+      [UTM.MEDIUM]: request.query[UTM.MEDIUM],
+      [UTM.CONTENT]: request.query[UTM.CONTENT],
+      [UTM.SOURCE]: request.query[UTM.SOURCE],
+      [UTM.TERM]: request.query[UTM.TERM]
+    }
+  })
+  return h.redirect(LICENCE_LENGTH.uri)
```
sgardnerdell, line 21: Shouldn't redirect to licence length - this will need to be configurable as an environment variable.

In production this will link to the external gov.uk landing page.

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -26,6 +27,21 @@ const getBackReference = async (request) => {
   return status.backRef.previous
 }
 
+const performTracking = async (request, path) => {
+  if ([CONTACT_SUMMARY.uri, LICENCE_SUMMARY.uri].includes(path)) {
```
sgardnerdell, line 31: Should only send a product view when on the licence summary, not on the contact summary

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -26,6 +27,21 @@ const getBackReference = async (request) => {
...
+    const transaction = await request.cache().helpers.transaction.get()
+    request.ga.ecommerce().detail(
```
sgardnerdell, line 33: Missing await on this call

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -26,6 +27,21 @@ const getBackReference = async (request) => {
...
+      getTrackingProductDetailsFromTransaction(transaction)
+    )
+  }
+  if (path === TERMS_AND_CONDITIONS.uri) {
+    const transaction = await request.cache().helpers.transaction.get()
+    request.ga.ecommerce().add(
```
sgardnerdell, line 39: Missing await on this call

packages/gafl-webapp-service/src/handlers/agreed-handler.js
```diff
@@ -142,6 +150,10 @@ const processPayment = async (request, transaction, status) => {
     // Defer setting the completed status in the journal until after finalization
     status[COMPLETION_STATUS.paymentCompleted] = true
     await request.cache().helpers.status.set(status)
+
+    request.ga.ecommerce().purchase(
```
sgardnerdell, line 154: Missing await on this call

packages/gafl-webapp-service/src/handlers/agreed-handler.js
```diff
@@ -73,6 +74,13 @@ const createPayment = async (request, transaction, status) => {
    */
   const paymentResponse = await sendPayment(preparedPayment)
 
+  /*
+   * Google Analytics tracking
+   */
+  request.ga.ecommerce().checkout(
```
sgardnerdell, line 80: Missing await on this call

packages/gafl-webapp-service/src/server.js
```diff
@@ -60,34 +62,74 @@ const createServer = options => {
...
+          {
+            id: process.env.ANALYTICS_PRIMARY_PROPERTY,
+            hitTypes: ['pageview', 'event', 'ecommerce']
+          },
+          {
+            id: process.env.ANALYTICS_EXGOV_PROPERTY,
+            hitTypes: ['pageview']
+          }
+        ],
+        sessionIdProducer: request => useSessionCookie(request) ? request.state[getSessionCookieName()].id : null,
```
graham-willis-druid, line 109: It might be neater to expose the id on the request object: add a getId() function to the cache decorator return object and simply reference it ...
```suggestion
        sessionIdProducer: request => request.cache().getId
```
and then you don't need useSessionCookie() in server.

packages/gafl-webapp-service/src/session-cache/session-manager.js
```diff
@@ -30,6 +30,8 @@ const protectionExemptSet = [
 
 const forbiddenUnlessAgreedSet = [ORDER_COMPLETE.uri, ORDER_COMPLETE_PDF.uri, PAYMENT_FAILED.uri, PAYMENT_CANCELLED.uri]
 
+export const useSessionCookie = request => request.path.startsWith('/buy')
```
sgardnerdell, line 33: My understanding was that we agreed to serve the session cookie for all pages which makes this redundant

packages/gafl-webapp-service/src/services/analytics/analytics.js
```diff
@@ -0,0 +1,15 @@
+
+export const getTrackingProductDetailsFromTransaction = transaction =>
+  transaction.permissions.map(permission => ({
+    id: permission.permit.description,
+    name: `${permission.permit.permitSubtype.label} - ${permission.permit.numberOfRods} rod(s) licence`,
+    brand: permission.permit.permitType.label,
+    category: [
+      permission.permit.permitSubtype.label,
+      `${permission.permit.numberOfRods} rod(s)`,
+      permission.permit.concessions.length ? permission.permit.concessions.join(',') : 'Full'
```
sgardnerdell, line 10: The concessions array is an array of object, not string.  We'll need to do a join on each concession object's name property:
```
[
  {
    "id": "d1ece997-ef65-e611-80dc-c4346bad4004",
    "name": "Disabled"
  }
]

File: packages/gafl-webapp-service/src/services/analytics/analytics.js [hunk 10]
```diff
@@ -0,0 +1,15 @@
+
+export const getTrackingProductDetailsFromTransaction = transaction =>
+  transaction.permissions.map(permission => ({
```
sgardnerdell, line 3: I know I didn't use this in the sample code but on reflection it might be nicer to use object destructing here to change from:
`transaction.permissions.map(permission => ({`
to
`transaction.permissions.map( { permit } => ({`

as all code below does permission.permit to access the permit object.

packages/gafl-webapp-service/src/server.js
```diff
@@ -60,34 +62,74 @@ const createServer = options => {
...
+                content: attribution[UTM.CONTENT],
+                medium: attribution[UTM.MEDIUM],
+                source: attribution[UTM.SOURCE],
+                term: attribution[UTM.TERM]
+              })
+            }
+          }
+          return {}
+        },
+        batchSize: 20,
```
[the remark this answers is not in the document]
sgardnerdell: These will be defaulted by hapi-gapi anyway so you could just remove if you'd prefer

packages/gafl-webapp-service/src/handlers/attribution-handler.js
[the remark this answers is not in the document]
sgardnerdell: I also added a comment suggesting to extract this to an environment variable

docker/env/gafl_webapp.env
```diff
@@ -24,3 +24,7 @@ SALES_API_URL=http://host.docker.internal:4000
 
 # Debug settings
 DEBUG=webapp:*
+
+# Tracking
+ANALYTICS_PRIMARY_PROPERTY=UA-124887477-9
```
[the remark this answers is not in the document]
sgardnerdell: These don't need to be kept secret but probably not required by default when running locally.

packages/gafl-webapp-service/src/session-cache/session-manager.js
jaucourt: Yes - this was what I was discussing with you and Haroon on the call on Monday. Moving over to that might best best extracted as a separate ticket, I think it'll take me another day to make that change.

graham-willis-druid: I think Phil was going to do that as a separate piece of work

packages/gafl-webapp-service/README.md
```diff
@@ -24,7 +24,8 @@ To run from this directory:
 | GOV_PAY_APIKEY             | GOV pay access identifier                       |    no    | Yes                 |                               |
 | GOV_PAY_REQUEST_TIMEOUT_MS | Timeout in milliseconds for API requests        |    no    | Yes                 |                               |
 | FEEDBACK_URI               | Location of feedback survey                     |    no    | #                   |                               |
-
+| ANALYTICS_PRIMARY_PROPERTY | Analytics ID for tracking inc ecommerce         |   yes    |                     |                               |
```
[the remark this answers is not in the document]
jaucourt: You won't be able to do a full run through, as if they're not set then the Hapi Gapi plugin isn't initialised and request.ga will be undefined. I can obviously check that so that it doesn't break, but then errors would end up hidden. Perhaps a log when Hapi Gapi doesn't get initialised, so we know that tracking isn't taking place

packages/gafl-webapp-service/src/services/analytics/analytics.js
```diff
@@ -0,0 +1,15 @@
+
```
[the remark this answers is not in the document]
jaucourt: Fair enough, I was a little unsure where to put a helper like this

packages/gafl-webapp-service/README.md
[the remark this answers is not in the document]
sgardnerdell: Good point Phil.   This might be something that we should improve in hapi gapi itself.

Perhaps rather than not initialising the plugin if these aren't set, only add the entries to the propertySettings if the individual variables are set:
```
        propertySettings: [
          {
            id: process.env.ANALYTICS_PRIMARY_PROPERTY,
            hitTypes: ['pageview', 'event', 'ecommerce']
          },
          {
            id: process.env.ANALYTICS_EXGOV_PROPERTY,
            hitTypes: ['pageview']
          }
        ],
```

packages/gafl-webapp-service/src/session-cache/session-manager.js
sgardnerdell: 👍

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
```diff
@@ -25,7 +26,7 @@ const checkPageCompleted = (p, status) => {
   }
 }
 
-const getData = async request => {
+export const getData = async request => {
```
graham-willis-druid, line 29: Q. Why did you export getData()

packages/gafl-webapp-service/src/pages/terms-and-conditions/route.js
```diff
@@ -1,11 +1,12 @@
 import pageRoute from '../../routes/page-route.js'
 import * as mappings from '../../processors/mapping-constants.js'
+import { getTrackingProductDetailsFromTransaction } from '../../processors/analytics.js'
 import Joi from '@hapi/joi'
 import { TERMS_AND_CONDITIONS, CONTROLLER, CONTACT_SUMMARY, LICENCE_SUMMARY } from '../../uri.js'
 
 import GetDataRedirect from '../../handlers/get-data-redirect.js'
 
-const getData = async request => {
+export const getData = async request => {
```
graham-willis-druid, line 9: Same question - why the export

packages/gafl-webapp-service/src/server.js
```diff
@@ -61,34 +63,79 @@ const createServer = options => {
 // This is a hash of the inline script at line 31 of the GDS template. It is added to the CSP to except the in-line
 // script. It needs the quotes.
 const scriptHash = "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='"
-const plugIns = [
```
graham-willis-druid, line 64: This is getting rather cumbersome - we ought to think about extracting it to another file at some point - not now

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
jaucourt: To test it in isolation

packages/gafl-webapp-service/src/pages/terms-and-conditions/route.js
jaucourt: Same answer. Whilst it might be possible to extract it from the default export, it's easier and more transparent to just export it and test it as a unit.

graham-willis-druid: yes sure

PR #203, opened 2020-06-26
Add an NI number validator

graham-willis-druid:
Add an NI number validator to the business rules and use it on the current NI page

### DIFF
--- a/packages/business-rules-lib/src/validators/contact.js
+++ b/packages/business-rules-lib/src/validators/contact.js
@@ -286,3 +286,25 @@ export const createLastNameValidator = joi =>
     .external(capitaliseNamePrefixes(['Mc', "O'"]))
     .required()
     .example('Tester')
+
+/**
+ * Create a validator to check a valid national insurance number
+ *
+ * @param {Joi.Root} joi the joi validator used by the consuming project
+ * @returns {Joi.AnySchema}
+ */
+
+const ukNINORegEx = /^([ABCEGHJ-PRSTW-Z][ABCEGHJ-NPRSTW-Z])\s?([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([ABCD])$/
+const ukNINORegEx2 = /^(?!(BG|GB|KN|NK|NT|TN|ZZ))/
+
+export const createNationalInsuranceNumberValidator = joi =>
+  joi
+    .string()
+    .trim()
+    .uppercase()
+    .pattern(ukNINORegEx)
+    .pattern(ukNINORegEx2)
+    .replace(ukNINORegEx, '$1 $2 $3 $4 $5')
+    .required()
+    .description('A UK national insurance number')
+    .example('NH 12 34 56 A')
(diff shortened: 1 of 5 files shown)


packages/business-rules-lib/src/validators/contact.js
```diff
@@ -286,3 +286,25 @@ export const createLastNameValidator = joi =>
...
     .example('Tester')
+
+/**
+ * Create a validator to check a valid national insurance number
+ *
+ * @param {Joi.Root} joi the joi validator used by the consuming project
+ * @returns {Joi.AnySchema}
+ */
+
+const ukNINORegEx = /^([ABCEGHJ-PRSTW-Z][ABCEGHJ-NPRSTW-Z])\s?([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([ABCD])$/
```
jaucourt, line 297: I'm not 100% this is correct. My NI number starts 'JL', but looks like the L isn't a valid second character

sgardnerdell, line 297: These regular expressions should be combined into a single expression for efficiency.  A negative lookbehind can be used to accomplish this.  I haven't tested it but something like this should work:

`const ukNINORegEx = /^([ABCEGHJ-PRSTW-Z][ABCEGHJ-NPRSTW-Z])(?<!(?BG|GB|KN|NK|NT|TN|ZZ))\s?([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([0-9]{2})\s{0,3}([ABCD])$/`

packages/gafl-webapp-service/src/pages/concessions/benefit-ni-number/benefit-ni-number.njk
```diff
@@ -6,21 +6,20 @@
...
     <p class="govuk-body-m">We use this information to find out if you are eligible for a disabled concession</p>
-
     {{ govukInput({
       label: {
         text: "National Insurance number"
       },
       id: "ni-number",
       name: "ni-number",
-      hint: { html: "<p>You can find this on your payslip, P60 or tax return</p><p>For example, ‘QQ 12 34 56 C’</p>" },
+      hint: { html: "<p>You can find this on your payslip, P60 or tax return</p><p>For example, ‘NH 12 34 56 C’</p>" },
```
sgardnerdell, line 22: GDS guidance states to use QQ 12 34 56 C as the example because it is not valid.

packages/business-rules-lib/src/validators/contact.js
sgardnerdell: The expression does allow L as this is part of the J-N range defined as allowed for the second character

graham-willis-druid: Done

packages/gafl-webapp-service/src/pages/concessions/benefit-ni-number/benefit-ni-number.njk
graham-willis-druid: done

PR #236, opened 2020-07-13
Add support to authenticate with the configured redis instance

sgardnerdell:
(also contains some linting updates which should have been trapped by prettier-standard)


packages/gafl-webapp-service/src/handlers/attribution-handler.js
```diff
@@ -1,6 +1,6 @@
 import { UTM, ATTRIBUTION_REDIRECT_DEFAULT } from '../constants.js'
 
-const sanitiseInput = value => value ? encodeURIComponent(value) : value
+const sanitiseInput = value => (value ? encodeURIComponent(value) : value)
```
graham-willis-druid, line 3: To be replaced by disinfect. I will do that

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -17,7 +17,7 @@ export const errorShimm = e => e.details.reduce((a, c) => ({ ...a, [c.path[0]]:
  * @param pageData
  * @returns {Promise<void>}
  */
-const getBackReference = async (request) => {
+const getBackReference = async request => {
```
graham-willis-druid, line 20: To be redesigned. I will do that.

packages/gafl-webapp-service/src/pages/order-complete/__tests__/order-complete.spec.js
```diff
@@ -41,11 +41,9 @@ describe('The order completion handler', () => {
   it('throw a status 403 (forbidden) exception if the finalized flag is not set', async () => {
     await JUNIOR_12_MONTH_LICENCE.setup()
 
-    salesApi.createTransaction = jest.fn(async () =>
-      new Promise(resolve => resolve(JUNIOR_12_MONTH_LICENCE.transActionResponse)))
+    salesApi.createTransaction = jest.fn(async () => new Promise(resolve => resolve(JUNIOR_12_MONTH_LICENCE.transActionResponse)))
```
graham-willis-druid, line 44: Why are you concatenating lines to make them too long!

graham-willis-druid: Serious point tho, I am could be concurrently working on this now and it would throw up a lot of 'merge' noise

sgardnerdell: As mentioned in the comments, these were picked up by prettier-standard.

I'd like to understand how these aren't being formatted in the first place.   Prettier-standard should be picking these up.  The only way I think these could be getting in are if commits are being made with the --no-verify flag.

I'll look into a pre-push flag to deny the push if code is not properly formatted/linted

PR #260, opened 2020-07-27
Adjust licence validator, introduce disinfect library, refactoring improvements

graham-willis-druid:
(1) Changed the ordering of the messages on the renewals page.
(2) Relax validation of reference number to allow old licence numbers to be renewed
(3) Add input sanitisation - removed existing function around the attribution handler - 
this; http://0.0.0.0:3000/buy/attribution?utm_campaign='<script>HACK!!!</script>'&utm_medium='This is ok'
Is safe, as are inputs via the payload.
Note that disinfect simply removes anything that could be a script; <, > etc is escaped but  <.*> is replaced by ''. 
(4) Extracted the plugins to another file - server.js was getting too long

### DIFF
--- a/packages/business-rules-lib/src/validators/permission.js
+++ b/packages/business-rules-lib/src/validators/permission.js
@@ -11,6 +11,7 @@ export const permissionNumberUniqueComponentValidator = joi =>
   joi
     .string()
     .trim()
+    .uppercase()
     .pattern(/^[A-Z0-9]{6}$/)
     .required()
     .description('The unqiue part of the permission reference number')
--- a/packages/business-rules-lib/src/validators/permission.js
+++ b/packages/business-rules-lib/src/validators/permission.js
@@ -11,7 +11,7 @@ export const permissionNumberUniqueComponentValidator = joi =>
   joi
     .string()
     .trim()
-    .pattern(/^[A-HJ-NP-Z0-9]{6}$/)
+    .pattern(/^[A-Z0-9]{6}$/)
     .required()
     .description('The unqiue part of the permission reference number')
     .example('U6HLG9')
--- a/packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
@@ -27,20 +27,6 @@ describe('The attribution handler', () => {
     })
   })
 
-  it.each([
-    [UTM.CAMPAIGN, '<script>alert("hacked")</script>', '%3Cscript%3Ealert(%22hacked%22)%3C%2Fscript%3E'],
-    [UTM.MEDIUM, '<script>alert("busted")</script>', '%3Cscript%3Ealert(%22busted%22)%3C%2Fscript%3E'],
-    [UTM.CONTENT, '<script>alert("stolen")</script>', '%3Cscript%3Ealert(%22stolen%22)%3C%2Fscript%3E'],
-    [UTM.SOURCE, '<script>alert("pilfered")</script>', '%3Cscript%3Ealert(%22pilfered%22)%3C%2Fscript%3E'],
-    [UTM.TERM, '<script>alert("pinched")</script>', '%3Cscript%3Ealert(%22pinched%22)%3C%2Fscript%3E']
-  ])('Sanitises input values before persisting them', async (utmValue, sampleValue, sanitised) => {
-    const request = generateRequestMock({ [utmValue]: sampleValue })
-    await attributionHandler(request, generateResponseToolkitMock())
-    expect(request.cache.mock.results[0].value.helpers.status.set).toHaveBeenCalledWith({
-      attribution: expect.objectContaining({ [utmValue]: sanitised })
-    })
-  })
-
   it("redirects to ATTRIBUTION_REDIRECT_DEFAULT if ATTRIBUTION_REDIRECT env var isn't set", async () => {
     delete process.env.ATTRIBUTION_REDIRECT
     const query = { [UTM.CAMPAIGN]: 'campaign', [UTM.MEDIUM]: 'popup' }
--- a/packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
+++ b/packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
@@ -14,9 +14,9 @@
     set errorMap = {
         'referenceNumber' : {
             'string.empty': { text: 'Enter the last six digits of your licence number', ref: '#ref' },
-            'string.pattern.base': { text: 'The last six digits of your licence number doesn\'t look right. Check and enter again', ref: '#ref' },
+            'string.pattern.base': { text: 'The last six digits of your licence number don\'t look right. Check and enter again', ref: '#ref' },
             'string.invalid': {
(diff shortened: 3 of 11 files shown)


packages/business-rules-lib/src/validators/permission.js
```diff
@@ -11,7 +11,7 @@ export const permissionNumberUniqueComponentValidator = joi =>
   joi
     .string()
     .trim()
-    .pattern(/^[A-HJ-NP-Z0-9]{6}$/)
+    .pattern(/^[A-Z0-9]{6}$/)
```
sgardnerdell, line 14: This is requiring uppercase

packages/gafl-webapp-service/src/plugins.js
```diff
@@ -0,0 +1,93 @@
...
+import Crumb from '@hapi/crumb'
+import HapiGapi from '@defra/hapi-gapi'
+import { useSessionCookie } from './session-cache/session-manager.js'
+import { UTM } from './constants.js'
+
+// This is a hash of the inline script at line 31 of the GDS template. It is added to the CSP to except the in-line
+// script. It needs the quotes.
+const scriptHash = "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='"
+
+export const getPlugIns = (getSessionCookieName, getCsrfTokenCookieName) => {
```
sgardnerdell, line 15: getSessionCookieName & getCsrfTokenCookieName should be exported elsewhere and added as an import here rather than passing references.

Plugin is normally treated as one word so method should be getPlugins()

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
```diff
@@ -27,20 +27,6 @@ describe('The attribution handler', () => {
     })
   })
 
-  it.each([
```
sgardnerdell, line 30: Not sure why these tests have been removed?

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
```diff
@@ -12,6 +12,14 @@
 
 {%
     set errorMap = {
+        'referenceNumber' : {
+            'string.empty': { text: 'Enter the last six digits of your licence number', ref: '#ref' },
+            'string.pattern.base': { text: 'The last six digits of your licence number doesn\'t look right. Check and enter again', ref: '#ref' },
+            'string.invalid': {
+                text: 'The details you entered do not match those we have on record for licence number ' + payload.referenceNumber + '. Please check you have entered the correct date of birth and postcode',
```
sgardnerdell, line 19: This message really needs changing - it implies something that isn't true

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
graham-willis-druid: Functionality replaced by disinfect.

packages/gafl-webapp-service/src/plugins.js
graham-willis-druid: That is a matter of opinion not fact and as such I will not do it as an independent contractor I do the job according to my judgement; you really need to reduce the scope of your comments to those which have a material importance.

packages/business-rules-lib/src/validators/permission.js
graham-willis-druid: done

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
graham-willis-druid: Corrected grammar and spelling otherwise I can see nothing wrong with it.

packages/gafl-webapp-service/src/plugins.js
graham-willis-druid: For a quiet lift I have refactored this, but I think you should confine your comment to that which is materially significant. In the end this is a matter of taste - as the plugin file is really just an extension of the server file rather than being a separate logical module it is perfectly reasonable to pass a function reference which is primarily used in the main module file - in fact it serves to illustrate the relationship between the files. As an independent contractor I reserve the right to make these sort of decisions as to my own judgement.

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
sgardnerdell: > The details you entered do not match those we have on record for licence number ' + payload.referenceNumber + '. Please check you have entered the correct date of birth and postcode

This implies that we know that the licence number exists and that the dob/postcode is incorrect which is not the case.

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
sgardnerdell: But there are no tests to show that Disinfect is performing the required functionality

packages/gafl-webapp-service/src/plugins.js
sgardnerdell: Isn't the purpose of a review to seek the opinion of others?

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
graham-willis-druid: It does work. Why should I have to build tests for 3rd party libraries, its not my code - it is part of hapi.

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
graham-willis-druid: Yes it was developed initially for the link on the email so the licence number will be found. In the case of the letter where the user types it does not do anything different.

graham-willis-druid: I would have to alter the error messaging to know which route it took. Right now I am not sure this is a high priority

sgardnerdell: Surely both routes will not check the licence number existence in isolation of the other fields?  Otherwise this is a security issue.

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
sgardnerdell: I guess the purpose would be to check that the integration is working as expected.  Was just a little surprised to see existing tests removed.

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
graham-willis-druid: No its just the message - once structural validation has occurred the the items are checked as a tuple. The message was written for the email link which contains the licence number, so its was fair to assume that the licence number was correct and the message reflected that. It has since been changed to allow for the letter (generic) renewal so the error message was no longer appropriate. I have simplified the message now so it should apply to both

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
graham-willis-druid: Phil has the handlers directly mocked so his tests could not be adapted. I have now added tests to make sure the sanitizer works for all vectors

PR #261, opened 2020-07-27
Fix issues in SQS receiver and Sales API connector

sgardnerdell:
- Fix issue with SQS receiver which was unnecessarily imposing a delay on polling even if messages were received
- Ensure Sales API connector using the `Accept: application/json` header (AWS ALB was returning an HTML response)
- Revise logging for the POCL job and the Sales API to facilitate easier debugging

### DIFF
--- a/packages/pocl-job/src/staging/create-transactions.js
+++ b/packages/pocl-job/src/staging/create-transactions.js
@@ -82,6 +82,7 @@ const createTransactionsInSalesApi = async (filename, state) => {
         record.stage = RECORD_STAGE.TransactionCreationFailed
         record.createTransactionError = apiResponse
         state.failed++
+        debug('Failed to create transaction for record: %o', record)
         await salesApi.createStagingException({
           transactionFileException: {
             name: `${filename}: FAILED-CREATE-${record.id}`,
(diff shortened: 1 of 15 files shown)


packages/pocl-job/src/staging/create-transactions.js
```diff
@@ -82,6 +82,7 @@ const createTransactionsInSalesApi = async (filename, state) => {
         record.stage = RECORD_STAGE.TransactionCreationFailed
         record.createTransactionError = apiResponse
         state.failed++
+        debug('Failed to create transaction for record: %o', record)
```
jaucourt, line 85: Should this still be here?

packages/sqs-receiver-service/src/receiver.js
```diff
@@ -27,26 +23,20 @@ const receiver = async () => {
...
     messageLastReceived = Date.now()
   }
 
   await showQueueStatistics(env.URL)
 
-  // Invoke the poll delay only on a small number of messages processed
-  // if they are coming in thick-and-fast then read again immediately
-  if (!messages) {
+  // Invoke the poll delay only if no messages were received from the queue
+  if (!messages.length) {
```
jaucourt, line 37: Think this could just be an else on the `if (messages.length) {` above

sgardnerdell: Agreed, change incoming

PR #263, opened 2020-07-28
Prevent expletives from being generated in the licence number

sgardnerdell:
Remove any character that could be treated as a vowel (inc 0, 1, 4) from the list of allowable characters.
This leaves 26 possible characters which is still an improvement on the previous 16 (hex) system.


packages/sales-api-service/src/services/permissions.service.js
```diff
@@ -38,8 +38,7 @@ export const generatePermissionNumber = async (
   const initials = (firstName.charAt(0) + lastName.charAt(0)).toUpperCase()
   const block2 = permit.numberOfRods + channel + type + duration + age + initials
 
-  const block3 = cryptoRandomString({ length: 6, characters: 'ABCDEFGHJKLMNPQRSTUVWXYZ0123456789' })
-
+  const block3 = cryptoRandomString({ length: 6, characters: 'BCDFGHJKLMNPQRSTVWXZ256789' })
```
graham-willis-druid, line 41: Is it worth tightening up the permission number validator to reflect this? The current validator describes a superset of the unique address space

sgardnerdell: Good point.  The API doesn't actually currently use the validator to actually do any validation - it is just used to form the swagger model to describe the response.  The current examples will now be technically invalid though so definitely something to look into.

sgardnerdell: I've updated the validators but have actually relaxed the permission number validators to allow for old-style permission numbers otherwise users would have been unable to renew licences.

PR #269, opened 2020-07-30
Performance tuning improvements

sgardnerdell:
- Address hotspots identified from performance profiling
- Increase default subscriber timeout in SQS receiver
- Ensure visibility timeout is greater than subscriber timeout in SQS receiver


packages/dynamics-lib/src/client/entity-manager.js
```diff
@@ -89,55 +85,44 @@ export function retrieveMultipleAsMap (...entityClasses) {
...
-      return data
-        .map((result, i) => result.value.map(v => entityClasses[i].fromResponse(v, optionSetData)))
-        .reduce((acc, val, idx) => {
-          acc[entityClasses[idx].definition.localCollection] = val
-          return acc
-        }, {})
+      return data.reduce((acc, item, idx) => {
+        acc[entityClasses[idx].definition.localCollection] = item.value.map(v => entityClasses[idx].fromResponse(v, optionSetData))
+        return acc
+      }, {})
```
jaucourt, line 91: I appreciate you're just changing this, but it's not the easiest thing in the world to make sense of - may benefit from a little more expressiveness, as I don't really understand what you're getting here. Perhaps just break over several lines, assigning to some well named variables for us ordinary mortals?

packages/dynamics-lib/src/client/entity-manager.js
```diff
@@ -89,55 +85,44 @@ export function retrieveMultipleAsMap (...entityClasses) {
...
-          optionSetData[name] = {
+        return data.value.reduce((acc, { Name: name, Options: options }) => {
+          acc[name] = {
             name,
             options: options.reduce((optionSetMapping, o) => {
-              optionSetMapping[o.id] = new GlobalOptionSetDefinition(name, o)
+              const id = o.Value
+              const label = o.Label?.UserLocalizedLabel?.Label
+              const description = o.Description?.UserLocalizedLabel?.Label || label
+              optionSetMapping[id] = { id, label, description }
```
jaucourt, line 114: would this be better as an `options.map` rather than `options.reduce`? Looks like we end up with an array of objects with remapped values at the end rather than a scalar value which is generally what `reduce` is for. Might've misunderstood though!

packages/dynamics-lib/src/entities/base.entity.js
```diff
@@ -163,8 +169,13 @@ export class BaseEntity {
    * @private
    */
   _setOptionSet (property, value) {
-    const mapping = this.constructor.definition.mappings[property]
-    if (!(value instanceof GlobalOptionSetDefinition) || value.optionSetName !== mapping.ref) {
+    if (
+      !(
+        Object.prototype.hasOwnProperty.call(value, 'id') &&
+        Object.prototype.hasOwnProperty.call(value, 'label') &&
+        Object.prototype.hasOwnProperty.call(value, 'description')
```
jaucourt, line 176: Could you not just do `value.hasOwnProperty('id') && value.hasOwnProperty('label') && value.hasOwnProperty('description')`?

Also, 
```
const validGlobalOptionSetDefinition = !(value.hasOwnProperty('id') && value.hasOwnProperty('label') && value.hasOwnProperty('description'))
if (validGlobalOptionSetDefinition) { //...
```
might be a bit clearer and more expressive

packages/dynamics-lib/src/entities/base.entity.js
```diff
@@ -263,66 +274,63 @@ export class BaseEntity {
...
+      },
+      Object.entries(this._bindings).reduce((acc, [k, v]) => {
+        if (v instanceof BaseEntity) {
+          acc[k] = v.id ? `/${v.constructor.definition.dynamicsCollection}(${v.id})` : `$${v.uniqueContentId}`
+        } else {
+          acc[k] = v
+        }
+        return acc
+      }, {})
+    )
```
jaucourt, line 301: This also needs a little more expressiveness to explain what's going on too - it's extremely terse, but quite difficult to work out what's going on (what's k? what's v?), perhaps break out into methods and consider using `map` again rather than reduce as it looks like it might be remapping rather than producing a scalar

packages/dynamics-lib/src/entities/base.entity.js
```diff
@@ -263,66 +274,63 @@ export class BaseEntity {
...
           }
         }
+      }
+      if (value !== undefined) {
         acc[property] = value
-        return acc
-      }, {})
+      }
+      return acc
+    }, {})
```
jaucourt, line 333: Again, a bit tricky to work out what's going on

packages/dynamics-lib/src/client/entity-manager.js
sgardnerdell: We can't use .map here as we're translating from an array into an object.

packages/dynamics-lib/src/entities/base.entity.js
sgardnerdell: Eslint warns about doing `value.hasOwnProperty('id') && value.hasOwnProperty('label') && value.hasOwnProperty('description')`

packages/dynamics-lib/src/entities/base.entity.js
sgardnerdell: Yes, agreed.  Some additional comments might help to describe what the method is doing

packages/dynamics-lib/src/entities/base.entity.js
sgardnerdell: Yes, agreed.  Some additional comments might help to describe what the method is doing.

Again, cannot use `.map` as we're translating from an array into an object

packages/dynamics-lib/src/client/entity-manager.js
sgardnerdell: Yes, agreed.  Some additional comments might help to describe what the method is doing

packages/dynamics-lib/src/entities/base.entity.js
jaucourt: Fair enough - have a think about my second comment on `const validGlobalOptionSetDefinition`

PR #307, opened 2020-08-13
Fix handling of date/times when running in docker container

sgardnerdell:
- Frontend was not converting permission UTC dates to local time for display
- POCL was not handling dates using local time before converting to UTC


packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -13,28 +13,36 @@ describe('displayStartTime', () => {
...
   })
-  it('displays the date in the correct format where the start time is 3pm', () => {
+  it('displays the date in the correct format where the start time is 3pm (UTC)', () => {
     const startTime = displayStartTime({ licenceStartDate: '2021-01-01', licenceStartTime: '15' })
     expect(startTime).toEqual('Friday, January 1st, 2021, 3:00pm')
   })
+  it('displays the date in the correct format where the start time is 3pm (BST)', () => {
+    const startTime = displayStartTime({ licenceStartDate: '2021-07-01', licenceStartTime: '15' })
+    expect(startTime).toEqual('Thursday, July 1st, 2021, 3:00pm')
+  })
```
jaucourt, line 23: Not sure what is differentiating this BST test from the UTC test above?

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -13,28 +13,36 @@ describe('displayStartTime', () => {
...
     const startTime = displayStartTime({ startDate: '2021-01-01T12:00:00.000Z', licenceStartDate: '2020-01-01', licenceStartTime: '12' })
     expect(startTime).toEqual('Friday, January 1st, 2021, 12:00pm (midday)')
   })
-  it('displays the date in the correct format where the API start time is 3pm', () => {
+  it('displays the date in the correct format where the API start time is 3pm (UTC)', () => {
     const startTime = displayStartTime({ startDate: '2021-01-01T15:00:00.000Z', licenceStartDate: '2021-01-01', licenceStartTime: '15' })
     expect(startTime).toEqual('Friday, January 1st, 2021, 3:00pm')
   })
+  it('displays the date in the correct format where the API start time is 3pm (BST)', () => {
+    const startTime = displayStartTime({ startDate: '2021-07-01T14:00:00.000Z', licenceStartDate: '2021-07-01', licenceStartTime: '15' })
```
jaucourt, line 43: If this is a BST date string, shouldn't it be `2021-07-01T14:00:00.000+01:00`?

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
sgardnerdell: It's because the date used falls into summertime (whereas the similar test above used Jan 1st as the date)

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
sgardnerdell: The Sales API and Dynamics always use UTC to represent dates.  Therefore they will always return `2021-07-01T14:00:00.000Z` rather than `2021-07-01T15:00:00.000+01:00`

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
jaucourt: https://www.youtube.com/watch?v=uuN7kOW7iwo

PR #312, opened 2020-08-17
OpenID Connect authentication and role based authorisation for telesales

sgardnerdell:


packages/gafl-webapp-service/src/handlers/oidc-handler.js
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * @param request the hapi request object
+ * @param h the hapi response handler
+ * @returns {Promise}
+ */
+export const signIn = async (request, h) => {
+  const success = !!request.payload.id_token
+  if (success) {
+    // Retrieve the nonce from the server cache for the given state value
+    const { nonce, state, postAuthRedirect = '/buy' } = (await cache.get(request.payload.state)) ?? {}
```
GrahamWillis, line 82: There is a constant for this route in uri.js  for this - CONTROLLER

packages/gafl-webapp-service/src/handlers/oidc-handler.js
```diff
@@ -0,0 +1,108 @@
...
+  db('Discovered issuer %s %O', MsIssuer.issuer, MsIssuer.metadata)
+
+  redirectUri = new URL(OIDC_SIGNIN.uri, process.env.OIDC_REDIRECT_HOST).toString()
+  client = new MsIssuer.Client({
+    client_id: process.env.OIDC_CLIENT_ID,
+    client_secret: process.env.OIDC_CLIENT_SECRET,
+    redirect_uris: [redirectUri],
+    response_types: ['code', 'id_token']
+  })
+  cache = server.cache({ segment: 'oidc', expiresIn: 12 * 60 * 60 * 1000 })
```
GrahamWillis, line 31: Should extract this to an environment variable? Security will try and get us to reduce it. -Should be synchronised with the ttl on the cookie in the same way as the session cookie. I think I would have integrated this with the existing session management.

packages/gafl-webapp-service/src/pages/oidc/role-required.njk
```diff
@@ -0,0 +1,23 @@
...
+        <h1 class="govuk-heading-l">Role required</h1>
+        <p class="govuk-body">
+            You were successfully signed in but you do not have the correct role assigned to access the telesales application.
+        </p>
+        <p class="govuk-body">
+            Please contact your Microsoft Dynamics administrator to request access.
+        </p>
+
+        {{ govukButton({
+            href: '/buy',
```
GrahamWillis, line 16: I have pushed the uri's down onto the pages so we can relocate anything easily - no hardcoded uri's.

packages/gafl-webapp-service/src/pages/oidc/account-disabled.njk
```diff
@@ -0,0 +1,23 @@
...
+        <h1 class="govuk-heading-l">Account disabled</h1>
+        <p class="govuk-body">
+            You were successfully signed in but your account is disabled in Microsoft Dynamics.
+        </p>
+        <p class="govuk-body">
+            Please contact your Microsoft Dynamics administrator to request access.
+        </p>
+
+        {{ govukButton({
+            href: '/buy',
```
GrahamWillis, line 16: I have pushed the uri's down onto the pages so we can relocate anything easily - no hardcoded uri's.

packages/gafl-webapp-service/src/handlers/oidc-handler.js
sgardnerdell: Refactored to use CONTROLLER.uri

packages/gafl-webapp-service/src/pages/oidc/account-disabled.njk
sgardnerdell: Refactored to use CONTROLLER.uri

packages/gafl-webapp-service/src/pages/oidc/role-required.njk
sgardnerdell: Refactored to use CONTROLLER.uri

packages/gafl-webapp-service/src/handlers/oidc-handler.js
sgardnerdell: This is a server-side cache which is used to store data while the user has been redirected to the OpenID connect endpoint, it is not the lifetime of the authenticated OIDC session (that is set below by examining the exp claim of the id_token returned when successfully authenticated).

This cache merely stores a randomly generated state and nonce value along with the page url requested before the authentication redirect occurred.  I know 12 hours seems a little arbitrary but this seemed like a reasonable value as the data stored here is very small and we don't want telesales users encountering an error if they sit on authentication page for several hours and then later enter the username/password.

PR #317, opened 2020-08-24
Security headers

graham-willis-druid:
Added headers as a result of the OWASP ZAP Scan


sgardnerdell (approved):

packages/gafl-webapp-service/src/session-cache/session-manager.js
```diff
@@ -30,8 +30,11 @@ const protectionExemptSet = [
 
 const forbiddenUnlessAgreedSet = [ORDER_COMPLETE.uri, ORDER_COMPLETE_PDF.uri, PAYMENT_FAILED.uri, PAYMENT_CANCELLED.uri]
 
-export const useSessionCookie = request =>
-  !request.path.startsWith('/public') && !request.path.startsWith('/oidc') && request.path !== '/robots.txt'
+const staticMatcherPublic = /^(\/public\/.*|^\/robots.txt)/
```
jaucourt, line 33: The start of string matcher needs to be in the bracket, and I think an end-of-string matcher would help on the robots.txt: `^\/(public\/|robots.txt$)` (check them here: https://regex101.com)

PR #324, opened 2020-08-26
Add encryption to session cookie

graham-willis-druid:


packages/gafl-webapp-service/README.md
```diff
@@ -17,6 +17,7 @@ To run from this directory:
 | REDIS_PASSWORD             | Password used to authenticate with the configured redis instance |    no    |                     |                               |
 | CHANNEL                    | The sales channel                                                |    no    | websales            | websales, telesales           |
 | SESSION_COOKIE_NAME        | Name of the session cookie                                       |    no    | sid                 |                               |
+| SESSION_COOKIE_PASSWORD    | Encryption key for the session cookie                            |   yes    |                     |                               |
```
jaucourt, line 20: Perhaps include the detail about it needing to be a 32-character string

packages/gafl-webapp-service/src/__mocks__/test-utils-system.js
```diff
@@ -10,6 +10,8 @@ import mockConcessions from './data/concessions'
 import mockDefraCountries from './data/defra-country'
 import { pricingDetail } from '../processors/pricing-summary'
 
+process.env.SESSION_COOKIE_PASSWORD = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
```
jaucourt, line 13: So that this is a bit clearer, you should do `("A").repeat(32)`

jaucourt (changes requested):

packages/gafl-webapp-service/README.md
graham-willis-druid: ok

packages/gafl-webapp-service/src/__mocks__/test-utils-system.js
graham-willis-druid: done

sgardnerdell (changes requested):
Changes look good but does @hapi/iron need to be added to the dependencies?

Will need a ticket on the backlog to deploy the changes into the environments - will have to hold off until after the pen test is completed.

graham-willis-druid:
> Changes look good but does @hapi/iron need to be added to the dependencies?
> 
> Will need a ticket on the backlog to deploy the changes into the environments - will have to hold off until after the pen test is completed.

I didn't need to add iron - I think its built into hapi and is only needed as a package if you want to use it independently of hapi.

I noticed that your cookie does not use it...{
    cookie: {
      name: process.env.OIDC_SESSION_COOKIE_NAME,
      password: process.env.OIDC_SESSION_COOKIE_PASSWORD,
      ttl: null,
      isSecure: process.env.NODE_ENV !== 'development',
      isHttpOnly: process.env.NODE_ENV !== 'development',
      isSameSite: 'Lax',
      path: '/'
    }
That is using the default encryption of none. Do you want me to add it in this PR?

sgardnerdell:
No need to change the OIDC cookie - its the default for cookies set via @hapi/cookie.

sgardnerdell (approved):

packages/gafl-webapp-service/README.md
sgardnerdell: Re-reviewed and this has been actioned

packages/gafl-webapp-service/src/__mocks__/test-utils-system.js
sgardnerdell: Re-reviewed and this has been actioned

PR #325, opened 2020-08-26
Fix OIDC handler to handle accounts not present in Dynamics

sgardnerdell:


packages/gafl-webapp-service/src/handlers/oidc-handler.js
```diff
@@ -87,9 +87,9 @@ export const signIn = async (request, h) => {
     const { oid, name, email, exp } = tokenSet.claims()
 
     const userDetails = await salesApi.getSystemUser(oid)
-    const hasTelesalesRole = !!userDetails.roles.find(role => role.name === process.env.OIDC_REQUIRE_DYNAMICS_ROLE)
+    const hasTelesalesRole = !!userDetails?.roles.find(role => role.name === process.env.OIDC_REQUIRE_DYNAMICS_ROLE)
```
graham-willis-druid, line 90: Yes I like the ?. Must use it soon.

graham-willis-druid (approved):

PR #345, opened 2020-09-09
Multiple frontend functional and content changes

graham-willis-druid:
(1) SIGINT SIGTERM Handling.

(2) Mobile phone validation added. This is quite strict and will only allow UK mobile numbers. The GDS guidance is clear that we should not gather numbers unless we have a good use for them and as we don't send SMS outside the UK we should not collect them. This should also help get a cleaner data set. I have allowed +447... and 07... This is done in the f/e so as not to impact the POCL loader.   

(3) Some minor changes to the printout - 1373. We may rebrand this to EA rather than DEFRA at a later stage - tbc.

(4)  Added the HSTS header as per the PEN test. I have confirmed with Tom that we don't register out domains for HSTS preloading so I have used the basic version of the header described here: [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security]

(5) Changes to the licence length impact on the contact details so capture this in a generic onLicenceLengthChange method

(6) Cookie / cache timeout always result in the journey going back to the beginning.

(7) With a disabled concession changing from 1 12 month licence to a 1 day restores the previously added concession.

(8) Multiple minor content tweaks.

(9) Length changes to the 3 rod licence are now allowed. Reducing the length of the 3 rod licence sets it to the two rod.

(10) Empty error fixed on the newsletter page.


packages/gafl-webapp-service/src/pages/licence-details/licence-length/update-transaction.js
```diff
@@ -1,7 +1,71 @@
 import * as mappings from '../../../processors/mapping-constants.js'
 import { LICENCE_LENGTH } from '../../../uri.js'
 import * as concessionHelper from '../../../processors/concession-helper.js'
-import { checkAfterPayment } from '../licence-to-start/update-transaction.js'
+import moment from 'moment'
```
sgardnerdell, line 4: Should this be moment-timezone?

graham-willis-druid: Yes changed

sgardnerdell (approved):

PR #358, opened 2020-09-15
Fix calendar noshow on IE11

graham-willis-druid:
Calendar does not work on IE due to div:content different behaviour. Change to use standard image.


sgardnerdell (commented):
Would prefer to use the background-image css property (which is definitely supported under IE11):
```
background-image: url("/public/images/icon-calendar-2x.png")
```

sgardnerdell (approved):

PR #363, opened 2020-09-17
Revise contact name validation rules

sgardnerdell:
- Allow single character names at the API (for POCL compatibility)
- Tighten up regex name validator (would have allowed .. as a valid name)


packages/business-rules-lib/src/validators/contact.validators.js
```diff
@@ -241,7 +241,7 @@ const createNameStringValidator = joi =>
     rules: {
       allowable: {
         validate (value, helpers) {
-          if (!/^[\p{L}\s\-'.]*$/gu.test(value)) {
+          if (!/^\p{L}+(?:(?:\.?\s|[-'\s])\p{L}+)*$/u.test(value)) {
```
jaucourt, line 244: Fully appreciate you're just modifying what was already here, but I think there's a danger of unnecessary compilations by putting the regex here rather than declaring it outside of the function scope

jaucourt (commented):
Looks ok apart from the one comment I made (which wasn't really done by you!); however what was the driver to reduce the minimum length of contact name from 2 characters to 1?

packages/business-rules-lib/src/validators/contact.validators.js
sgardnerdell: I would have hoped that the V8 optimiser would compile the expression the first time it is encountered and then cache this result, but happy to move this to a variable and reference that.  Change incoming.

sgardnerdell:
> Looks ok apart from the one comment I made (which wasn't really done by you!); however what was the driver to reduce the minimum length of contact name from 2 characters to 1?

The Post Office counter system only enforces a minimum of 1 character and we're unable to get that changed before release.  I would hope in the longer term we could get their system to enforce a minimum of 2 characters.

graham-willis-druid (approved):

packages/business-rules-lib/src/validators/contact.validators.js
sgardnerdell: I wouldn't call this conclusive, but I knocked up a quick test for this:

`[1000, 10000, 100000, 1000000, 10000000, 1000000000].forEach((iterations) => {
  const nameStringRegex = /^\p{L}+(?:(?:\.?\s|[-'\s])\p{L}+)*$/u

  const testPrecompiled = (value) => {
    for (let i = 0; i < iterations; i++) {
      if (!nameStringRegex.test(value)) {
        throw new Error('Should never happen')
      }
    }
  }

  const testInlineCompilation = (value) => {
    for (let i = 0; i < iterations; i++) {
      if (!/^\p{L}+(?:(?:\.?\s|[-'\s])\p{L}+)*$/u.test(value)) {
        throw new Error('Should never happen')
      }
    }
  }

  console.time(`testPrecompiled for ${iterations} iterations`)
  testPrecompiled('Peter Paul')
  console.timeEnd(`testPrecompiled for ${iterations} iterations`)

  console.time(`testInlineCompilation for ${iterations} iterations`)
  testInlineCompilation('Peter Paul')
  console.timeEnd(`testInlineCompilation for ${iterations} iterations`)
})
`

The results are:
`testPrecompiled for 1000 iterations: 0.553ms
testInlineCompilation for 1000 iterations: 0.168ms
testPrecompiled for 10000 iterations: 0.774ms
testInlineCompilation for 10000 iterations: 1.575ms
testPrecompiled for 100000 iterations: 6.553ms
testInlineCompilation for 100000 iterations: 7.343ms
testPrecompiled for 1000000 iterations: 38.61ms
testInlineCompilation for 1000000 iterations: 38.573ms
testPrecompiled for 10000000 iterations: 359.211ms
testInlineCompilation for 10000000 iterations: 391.304ms
testPrecompiled for 1000000000 iterations: 35.633s
testInlineCompilation for 1000000000 iterations: 39.791s`

PR #367, opened 2020-09-18
Bugfix: Junior Licences not tracked as sales

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-1340

Junior licences do not get tracked as sales, most likely because they don't go through GovPay or have a 0 transaction value.


packages/gafl-webapp-service/src/handlers/agreed-handler.js
```diff
@@ -212,6 +212,10 @@ export default async (request, h) => {
     if (next) {
       return h.redirect(next)
     }
+  } else {
+    await request.ga
+      .ecommerce()
+      .purchase(getTrackingProductDetailsFromTransaction(transaction), '', getAffiliation(process.env.CHANNEL))
```
sgardnerdell, line 218: I'm pretty sure that the payment-id is a mandatory field in order for GA to record the transaction as an ecommerce goal.  In the old service we simply used the session id here.

jaucourt: doneski

sgardnerdell (approved):

PR #373, opened 2020-09-22
Rebrand PDF printout to be EA rather than GDS

graham-willis-druid:
The licence is issued by the EA, not GDS. The printout is branded accordingly


packages/gafl-webapp-service/src/processors/__tests__/confirmation-pdf.spec.js
```diff
@@ -40,8 +40,8 @@ const disabledPermission = {
 describe('The PDF generator', () => {
   it('completes for a disabled concession trout and coarse licence', () => {
     const result = orderConfirmationPdf(disabledPermission)
-    expect(result.content[5].table.body[1][1].text).toBe('Trout and coarse, up to 3 rods')
-    expect(result.content[5].table.body[3][1].text).toBe('yes')
+    expect(result.content[4].table.body[1][1].text).toBe('Trout and coarse, up to 3 rods')
```
jaucourt, line 43: What's special about index 4? Could you extract this to a variable with a descriptive name?

jaucourt (changes requested):
Just one small thing

packages/gafl-webapp-service/src/processors/__tests__/confirmation-pdf.spec.js
graham-willis-druid: Please see below

jaucourt (approved):

PR #374, opened 2020-09-22
Changes in view of Andrew Hick's test report; IWTF-1436

graham-willis-druid:
(1) When the address lookup fails to find any matching the manual address page now displays an inset message to this effect in the same way as the old service.
(2)  IE11 conditional expansion errors fixed
(3) On both the licence type page and the licence length page the pricing boxes are set to initially hidden which means that if Javascript is switched off they will never be displayed.
(4) The links are now consistently styled throughout the journey
(5) Addition of the OS T&Cs page from the link in the address finder, which was missing entirely! (Not on Andrews report)
(6) Added accessibility statement - modified the RCR statement. (This will need a review from content)
(7) Align visually hidden text with title text on the licence summary
(8) T&Cs page width reduced
(8) https://github.com/alphagov/govuk-design-system/pull/522 - novalidate added to all forms
(9) Various content changes


packages/gafl-webapp-service/src/routes/misc-routes.js
```diff
@@ -91,24 +98,13 @@ export default [
...
   {
     method: 'GET',
     path: ATTRIBUTION.uri,
     handler: attribution
-  }
+  },
+  simpleView(ACCESSIBILITY_STATEMENT),
+  simpleView(PRIVACY_POLICY),
+  simpleView(REFUND_POLICY),
+  simpleView(OS_TERMS)
```
jaucourt, line 109: nice, like this!

jaucourt (approved):

PR #380, opened 2020-09-24
Implement Cookie Banner

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-1211

Add a cookie banner that can be dismissed for 90 days


graham-willis-druid:
How does this get closed if there if Javascript is disabled?

graham-willis-druid (approved):

PR #398, opened 2020-10-02
Implement distributed locks to prevent concurrent execution of ETL tasks

sgardnerdell:
- Implement distributed lock mechanism in connectors-lib
- Integrate distributed locks into each of the ETL tasks
- Fix frontend tests causing jest to forcibly terminate


packages/connectors-lib/src/distributed-lock.js
```diff
@@ -0,0 +1,103 @@
...
+    do {
+      debug('Attempting to obtain lock for %s', this._lockName)
+      try {
+        this._lock = await this._redlock.lock(this._lockName, this._ttl)
+        elapsed = (new Date() - invocationTime) / 1000
+      } catch (e) {
+        // Ignored
+      }
+      // eslint-disable-next-line no-unmodified-loop-condition
+    } while (!this._lock && (maxWaitSeconds < 0 || elapsed < maxWaitSeconds))
```
graham-willis-druid, line 59: Are you hammering on the lock here? Should this have a wait in it, or is it built into the lock?

packages/gafl-webapp-service/src/__tests__/server.spec.js
```diff
@@ -40,23 +40,25 @@ describe('The server', () => {
...
           await init()
           shutdownBehavior()
-          jest.spyOn(process, 'exit').mockImplementation(async () => {
-            expect(true).toBeTruthy()
-            // Allows time for the server to shut down
-            await new Promise(resolve => setTimeout(resolve, 1000))
-            done()
-          })
-          process.emit(signal)
+          const serverStopSpy = jest.spyOn(server, 'stop').mockImplementation(jest.fn())
```
graham-willis-druid, line 51: This is odd, because I could not get this very code to work. It kept giving me timeout errors in the jests tests until I eliminated it. It ran ok on its own, but when running the whole suite it was failing, repeatedly. Keep an eye on this.

graham-willis-druid (commented):
So this is a little bit scary at this point in time. As the ETL processes are single instances, would it not have been simpler and safer for the processes simply to write a lock file? I don't understand why we would need a distributed mechanism. What happens if the process takes longer to run than the TTL on lock in REDIS?

packages/gafl-webapp-service/src/__tests__/server.spec.js
sgardnerdell: I think the timeout errors were occurring because the function wasn't previously calling `await server.stop()` after  `jest.restoreAllMocks()` isn't called, which I believe was also why I was seeing Jest forcibly terminate the tests in the frontend.

packages/connectors-lib/src/distributed-lock.js
sgardnerdell: Yes, redlock has an internal wait/attempt mechanism each time you attempt to acquire the lock.

sgardnerdell:
It does seem a little scary but I am more concerned about the possible effects of a process being run concurrently which would have an unknown impact.  We cannot use a lock file because AWS Fargate uses ephemeral storage and in any case the step functions will spin up a completely separate container instance.  Using a distributed lock is the only solution here - I did consider using DynamoDB but Redis/Redlock offers the simpler and more reliable solution.

graham-willis-druid (approved):

PR #412, opened 2020-10-08
Update permission number generator to use a sequence and add a checksum

sgardnerdell:
- Permission numbers now generated based on a sequence no stored in redis
- Last digit of permission is now a checksum based on luhn mod 10 algorithm

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "extends": "standard",
     "parser": "babel-eslint",
     "env": {
+      "es2020": true,
       "node": true,
       "jest": true
     }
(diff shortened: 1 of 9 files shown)


graham-willis-druid (approved):
I like that there is a check digit. We need to test this hard.

sgardnerdell:
Yeah - it is definitely good to get that in place now!  I'll be running the performance test suite against this on an extended soak.  FYI, the dictionary gives us 4,734,976 unique permutations in the last 5 digits so based on current service usage it should be ~5 years before it rolls back round to the start of the sequence.

jaucourt (approved):

PR #423, opened 2020-10-12
Use dates in local time for licence start date calculation/validation in the frontend.

graham-willis-druid:
All remaining usages of moment() to use the local time pattern


packages/gafl-webapp-service/src/processors/api-transaction.js
```diff
@@ -67,7 +67,9 @@ export const prepareApiFinalisationPayload = async request => {
   return {
     payment: {
       amount: transaction.cost,
-      timestamp: moment().toISOString(),
+      timestamp: moment()
+        .tz(SERVICE_LOCAL_TIME)
```
graham-willis-druid, line 71: That is not necessary is it

sgardnerdell (approved):

PR #430, opened 2020-10-14
Relax validation rules for Post Office sales

sgardnerdell:
- No longer checks concession proof for P.O. sales
- No longer requires postcode for P.O. sales

### DIFF
--- a/packages/sales-api-service/src/schema/__tests__/transaction.schema.spec.js
+++ b/packages/sales-api-service/src/schema/__tests__/transaction.schema.spec.js
@@ -19,7 +19,21 @@ describe('createTransactionSchema', () => {
   it('validates successfully when issueDate and startDate are null', async () => {
     const mockPayload = mockTransactionPayload()
     mockPayload.permissions.map(p => ({ ...p, issueDate: null, startDate: null }))
-    const result = await createTransactionSchema.validateAsync(mockTransactionPayload())
+    const result = await createTransactionSchema.validateAsync(mockPayload)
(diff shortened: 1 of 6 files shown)


packages/sales-api-service/src/schema/__tests__/transaction.schema.spec.js
```diff
@@ -19,7 +19,21 @@ describe('createTransactionSchema', () => {
   it('validates successfully when issueDate and startDate are null', async () => {
     const mockPayload = mockTransactionPayload()
     mockPayload.permissions.map(p => ({ ...p, issueDate: null, startDate: null }))
-    const result = await createTransactionSchema.validateAsync(mockTransactionPayload())
+    const result = await createTransactionSchema.validateAsync(mockPayload)
```
jaucourt, line 22: Are you doing this to check it doesn't throw? `expect(result).toBeInstanceOf(Object)` looks like it's testing that as a side effect, it'd be better doing `expect(async () => await createTransactionSchema.validateAsync(mockPayload)).not.toThrow()` if that's the case

jaucourt (changes requested):

packages/sales-api-service/src/schema/__tests__/transaction.schema.spec.js
sgardnerdell: When an object passes validation a new object is returned which is based on the source object with any transformations (eg. `trim()`) applied to it.

jaucourt (approved):

PR #434, opened 2020-10-15
Update to handle new POCL file format with newsletter fields

philhunt-defra:


sgardnerdell (changes requested):
The changes to the POCL look sensible and I think you've captured everything that needs doing.

I don't think the changes to the gafl-webapp-service should be included though - possibly unintentionally changed due to a global search/replace?

sgardnerdell:
FAO: Graham/Phil B - we shouldn't merge this until the final test file has been supplied by the P.O. and this feature has been tested with that.

PR #438, opened 2020-10-19
Added telesales new customer button (#438)

graham-willis-druid:
(1) Add telesales clear link
(2) Fix renewals time comparator to use local time.


sgardnerdell:
The changes all look good, I'm just wondering whether the telesales staff will understand what 'Clear' means.  They might expect it to clear the form rather than start again.  Have Jon/Cecile agreed the terminology?

sgardnerdell (approved):

PR #484, opened 2020-11-17
Add logs to diagnose create transaction failure

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-1569

From time to time, the call to salesApi.createTransaction fails saying
birthDate is before the minimum date. We've been unable to track down
what's causing this, so we need to add more logs to give us a bit more
information

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/agreed-handler.js
+++ b/packages/gafl-webapp-service/src/handlers/agreed-handler.js
@@ -30,7 +30,14 @@ const debug = db('webapp:agreed-handler')
  */
 const sendToSalesApi = async (request, transaction, status) => {
   const apiTransactionPayload = await prepareApiTransactionPayload(request)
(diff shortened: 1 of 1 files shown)


packages/gafl-webapp-service/src/handlers/agreed-handler.js
```diff
@@ -30,7 +30,14 @@ const debug = db('webapp:agreed-handler')
  */
 const sendToSalesApi = async (request, transaction, status) => {
   const apiTransactionPayload = await prepareApiTransactionPayload(request)
-  const response = await salesApi.createTransaction(apiTransactionPayload)
+  delete apiTransactionPayload.permissions[0].licensee.birthDate
```
philhunt-defra, line 33: Should this be in here?

philhunt-defra (approved):

nabeelamir-defra (dismissed):

nabeelamir-defra (approved):

PR #539, opened 2021-01-05
Accessibility issues on the a problem has occurred page

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-1718

Fix for accessibility issues on the a problem has occurred page

### DIFF
--- a/packages/gafl-webapp-service/src/pages/errors/client-error.njk
+++ b/packages/gafl-webapp-service/src/pages/errors/client-error.njk
@@ -2,7 +2,7 @@
 {% from "fieldset/macro.njk" import govukFieldset %}
 {% from "button/macro.njk" import govukButton %}
 
-{% block pageTitle %}Error - Your request could not be processed - GOV.UK{% endblock %}
+{% block pageTitle %}A problem has occurred{% endblock %}
 
 {% set title = 'A problem has occurred' %}
 {% if clientError.statusCode === 400 %}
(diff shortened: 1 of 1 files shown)


packages/gafl-webapp-service/src/pages/errors/client-error.njk
```diff
@@ -18,18 +18,14 @@
...
               legend: {
-                text: title,
-                classes: "govuk-fieldset__legend--m govuk-!-margin-bottom-3",
+                text: "A problem has occurred",
+                classes: "govuk-fieldset__legend--l govuk-!-margin-bottom-6",
                 isPageHeading: true
               }
             }) %}
-
+            <h2 class="govuk-heading-m">{{ title }}</h2>
```
jaucourt, line 28: This looks like we'll end up with 'A problem has occurred' repeated two or three times on the page - or am I wrong?

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/errors/client-error.njk
nabeelamir-defra: If the statusCode is not 400, 401, 403, 404 then yes, title would be 'A problem has occurred' and be repeated twice. However, this is existing behaviour.

nabeelamir-defra: The pageTitle block is the text shown in browser's title bar

jaucourt (approved):

PR #542, opened 2021-01-11
Only show address plural when there are multiple addresses

philhunt-defra:


jaucourt (approved):

nabeelamir-defra (approved):
Approved, but could you add the jira number in the description please

PR #554, opened 2021-01-18
Only process files in S3 from the last week

philhunt-defra:


packages/pocl-job/src/io/s3.js
```diff
@@ -47,6 +47,11 @@ export const refreshS3Metadata = async () => {
 
     console.log(`Processing ${filename}`)
 
+    if (moment(new Date(file.LastModified)) < moment().subtract(1, 'weeks')) {
+      console.log('Skipping file older that one week')
+      continue
+    }
+
```
jaucourt, line 54: This is ok, but I feel it might be a bit more obvious to pre-filter the fileList, something like:

```
const recentFiles = filesList.filter(f => moment(new Date(file.LastModified)) > moment().subtract(1, 'weeks'))

console.log(`Found ${filesList.length - recentFiles.length} file(s) older than 1 week`)

for (const file of recentFiles) {
...
```

nabeelamir-defra (approved):

PR #616, opened 2021-03-11
Fix how we obtain client tracking id

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-2023

How we obtain the client id for tracking seems to be based on a misapprehension. We need to obtain the client id for tracking
correctly, but also in a manner that fails gracefully in case:

1) I'm wrong about how _ga is formatted
2) The way that _ga is put together should ever change in the future

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/analytics.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/analytics.spec.js
@@ -38,7 +38,7 @@ describe('initialiseAnalyticsSessionData', () => {
     })
 
   it.each([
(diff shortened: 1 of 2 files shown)


philhunt-defra (approved):

nabeelamir-defra (approved):

packages/gafl-webapp-service/src/processors/__tests__/analytics.spec.js
```diff
@@ -1,20 +1,69 @@
...
+    try {
+      await initialiseAnalyticsSessionData(fakeRequest)
+    } catch (e) {
+      err = e
+    }
+    expect(err).toBeUndefined()
+  })
+
+  it.each([
+    '!@£$%^&*()œ∑´®†¥¨^øåß∂ƒ©˙∆˚', '¡€#¢∞§¶•ªø^¨¥©˙∆˚˙©†ƒ®', '1$2$345$678', '1,2,345,678'
```
nabeelamir-defra, line 57: Should this contain a test for 523510731.1602852296, just for completeness.

PR #629, opened 2021-03-23
Order Complete screen text for digital licence

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-1930

There is different text on the order complete page depending on whether the Angler has opted for a paperless licence or not

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,9 @@ name: build
 
 on:
   pull_request:
+    branches:
+      - develop
+      - master
   push:
     branches:
       - develop
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 name: build
 
 on:
-  pull_request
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -2,9 +2,6 @@ name: build
 
 on:
   pull_request:
```
jaucourt, line 4: rather than deleting -develop and -master, it'd be better to just add your feature branch. I'm not sure what the outcomes might be of the automated versioning script. It only does some bits for develop and master, but the whole script isn't wrapped in an if, so there might be some unpredictable results.

nabeelamir-defra: I see what you mean. There is something below which says run the new-version if: ${{ github.event_name == 'push' }} . But you're right, we don't really know the affects of it are. I think rather than messing with it in this branch I'll create a separate ticket for it. For now I'll just revert it.

jaucourt (approved):

PR #668, opened 2021-04-20
Bugfix: Remove UK from address entry country list

DGajewska:
IWTF-2094

* Removes UK from country list in address entry drop down
* Replaces UK in test data with England
* Removes hardcoding of countryCode when caching licensee data from CRM


nabeelamir-defra:
Code changes look fine to me, but I wasn't expecting changes to the package-lock.json. What node version are you using?

packages/gafl-webapp-service/src/__mocks__/data/defra-country.js
```diff
@@ -11,6 +11,11 @@ export default {
       label: 'United Kingdom',
       description: 'GB'
```
nabeelamir-defra, line 12: Should United Kingdom above now be removed?

packages/gafl-webapp-service/src/processors/refdata-helper.js
```diff
@@ -8,8 +8,9 @@ const optionProc = c =>
 
 const local = {}
 
-const fetch = async () =>
-  optionProc(await salesApi.countries.getAll())
+const fetch = async () =>{
+  // Returns all countries from Dynamics
+  return optionProc(await salesApi.countries.getAll())
     .sort((a, b) => {
       const order = ['GB-ENG', 'GB-WLS', 'GB-SCT', 'GB-NIR', 'GB']
```
jaucourt, line 15: GB should be removed from here now - and the test will need to be updated

packages/gafl-webapp-service/src/pages/contact/address/entry/route.js
```diff
@@ -18,11 +18,16 @@ const validator = Joi.object({
...
+  const options = await countries.getAll();
+  return options.filter(country => country.code !== 'GB')
+}
+
 export default pageRoute(ADDRESS_ENTRY.page, ADDRESS_ENTRY.uri, validator, nextPage, async request => {
   const { addresses, searchTerms } = await request.cache().helpers.addressLookup.getCurrentPermission()
   return {
     searchTerms: !addresses?.length && searchTerms ? searchTerms : null,
-    countries: await countries.getAll(),
+    countries: await getCountryDropDownOptions(),
```
jaucourt, line 30: Would be good to have a new test for this

jaucourt (changes requested):

packages/gafl-webapp-service/src/__mocks__/data/defra-country.js
DGajewska: This data is used by quite a few tests, so removing this would break other tests. More specifically, the address lookup sets the country to 'United Kingdom', so it needs to be available in the master list, or it breaks other parts of the code.

packages/gafl-webapp-service/src/processors/refdata-helper.js
DGajewska: Removing it here only causes the order to change, so 'United Kingdom' is moved to the top of the list, rather than below the individual countries, so I'm not sure that it matters whether it is in or not as we remove the option for the dropdown list anyway, so I'll remove it :)

packages/gafl-webapp-service/src/pages/contact/address/entry/route.js
DGajewska: Ah, yes, good catch. I think I got carried away with finishing this ticket. I'll add that in :)

packages/gafl-webapp-service/src/processors/refdata-helper.js
DGajewska: After our Slack chat, I have left this in

nabeelamir-defra (approved):

jaucourt (approved):

PR #686, opened 2021-05-07
Fix: Use serial number as Payment id for POCL transactions

DGajewska:
IWTF-2137

* Adds serial number to transaction payload in POCL transform stream
* Updates creation of Transaction Journal to use serial number for reference id when type is 'payment'
* Adds serialNumber to createTransaction schema

### DIFF
--- a/packages/pocl-job/src/staging/create-transactions.js
+++ b/packages/pocl-job/src/staging/create-transactions.js
@@ -72,7 +72,7 @@ const createTransactionsInSalesApi = async (filename, state) => {
   if (state.buffer.length) {
     debug({buffer: state.buffer[0].createTransactionPayload})
     const createResults = await salesApi.createTransactions(state.buffer.map(item => item.createTransactionPayload))
-    debug({createResults: createResults.response})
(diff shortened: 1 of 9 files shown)


packages/pocl-job/src/staging/create-transactions.js
```diff
@@ -71,7 +71,7 @@ const getInitialState = async filename => {
 const createTransactionsInSalesApi = async (filename, state) => {
   if (state.buffer.length) {
     const createResults = await salesApi.createTransactions(state.buffer.map(item => item.createTransactionPayload))
-
+
```
nabeelamir-defra, line 74: This extra space can be removed

nabeelamir-defra (approved):

packages/pocl-job/src/staging/create-transactions.js
DGajewska: I ran the `prettier-standard` script when I noticed this, but looking at it now, I didn't push my changes up! This has now been removed :)

PR #702, opened 2021-05-11
Footer links broken on some pages

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2139

The footer links are broken on the order complete and payment failed page

### DIFF
--- a/packages/gafl-webapp-service/src/pages/order-complete/route.js
+++ b/packages/gafl-webapp-service/src/pages/order-complete/route.js
@@ -29,6 +29,7 @@ const getData = async request => {
   }
 
   await request.cache().helpers.status.set({ [COMPLETION_STATUS.completed]: true })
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/pages/order-complete/route.js
```diff
@@ -29,6 +29,7 @@ const getData = async request => {
   }
 
   await request.cache().helpers.status.set({ [COMPLETION_STATUS.completed]: true })
+  await request.cache().helpers.status.setCurrentPermission({ currentPage: ORDER_COMPLETE.page })
```
jaucourt, line 32: I saw the new tests above, which is fine, however I'd like to see a direct test of this. If I was to delete this line would any tests fail?

jaucourt (changes requested):
I _think_ we'll need another test, but ok other than that

jaucourt (approved):

PR #713, opened 2021-05-21
Cannot navigate directly to easy renewal page

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-1877

Cannot navigate directly to easy renewal page


DGajewska (approved):

packages/gafl-webapp-service/src/session-cache/__tests__/session-manager.spec.js
```diff
@@ -82,4 +100,15 @@ describe('The user', () => {
     expect(response.statusCode).toBe(302)
     expect(response.headers.location).toBe(CONTROLLER.uri)
   })
+
+  it('returns 200 when attempting to access ' + IDENTIFY.uri, async () => {
```
jaucourt, line 104: You can use template quotes here rather than concatenation: `` `returns 200 when attempting to access ${IDENTIFY.url}` ``

jaucourt (changes requested):
One small thing, otherwise looks good

jaucourt (approved):

PR #731, opened 2021-06-02
Fix: Only pass street name to GovPay if provided

DGajewska:
IWTF-2170

* Adds conditonal logic to prefilled billing address
* Adds tests for payment.js

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
@@ -4,12 +4,10 @@ import { licenceTypeAndLengthDisplay } from '../licence-type-display.js'
 jest.mock('../licence-type-display.js');
 licenceTypeAndLengthDisplay.mockReturnValue('Trout and coarse, up to 2 rods, 8 day')
 
-const createRequest = () => ({
-  headers: {
-    'x-forwarded-proto': 'https'
-  },
-  info: { host: 'localhost:1234' },
-  server: { info: { protocol: 'http' } },
+const createRequest = (opts = {}) => ({
+  headers: opts.headers || { 'x-forwarded-proto': 'https' },
+  info: { host: opts.host || 'localhost:1234' },
+  server: { info: { protocol: opts.protocol || '' } },
 })
 
 const createTransaction = () => ({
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
```diff
@@ -0,0 +1,115 @@
...
+describe('preparePayment', () => {
+  let request, transaction, result
+  beforeEach(() => {
+    request = createRequest()
+    transaction = createTransaction()
+    result = preparePayment(request, transaction)
+  })
+  describe('provides the correct return url', () => {
+    it('when the "x-forwarded-proto" header is present', () => {
+      expect(result.return_url).toBe('https://localhost:1234/buy/agreed')
```
jaucourt, line 42: After looking for a bit, I can see that `result.return_url` should be the same as ` ``http://${request.info.host}/buy/agreed`` `, however a bit more of a thorough test would be a bit more descriptive. If you were to modify the `createRequest` method to allow for host to be specified, but default to `localhost:1234`, you could then do something like:
```
['localhost:4321', 'otherhost:8888', 'samplehost:4444'].forEach(host => 
  it(`uses SSL when 'x-forwarded-proto' header is present, host '${host}'`, () => {
    const request = createRequest(host)
    expect(result.return_url).toBe(`https://${host}/buy/agreed`)
  })
)
```

packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
```diff
@@ -0,0 +1,115 @@
...
+    })
+    
+    it('when the "x-forwarded-proto" header is not present', () => {
+      request.headers = {}
+      const result = preparePayment(request, transaction)
+      expect(result.return_url).toBe('http://localhost:1234/buy/agreed')
```
jaucourt, line 48: Same here, but use a different list to previously

jaucourt (changes requested):
It's great that you've added in quite a few new tests that cover more than the new code you've added, thanks for doing that. I've made a couple of suggestions to broaden tests to be a little more expressive, this could be done with most of your tests, as at the moment the tests could be passed by simply hardcoding values. This probably wouldn't ever happen, but it would make them slightly more expressive and easy to understand, as well as allowing edge cases to be tested.

nabeelamir-defra (approved):

jaucourt (approved):

PR #764, opened 2021-06-17
Initial multibuy functionality

DGajewska:
IWTF-2233

* Adds licence length to mapping constants and removes duplicate code
* Create 'add licence' page
* Create 'view licences' page
* Pulls date and time display changes from digital licence branch
* Adds 'Change' functionality on 'view licences' page


nabeelamir-defra (approved):

jaucourt (changes requested):
No issues with most of this, however we need to talk between the three of us about the tests in GAFL. I don't think we should be following the pattern established there; whilst we can't just replace them all wholesale, I think we need proper unit tests rather than just testing status codes on a whole stack of components and when we write new code, it should be tested with those type of unit tests.

packages/gafl-webapp-service/src/pages/multibuy/add-licence/__tests__/route.spec.js
```diff
@@ -0,0 +1,51 @@
...
+import { createMockRequest } from '../../../../__mocks__/request.js'
+
+describe('The add licence route .getData', () => {
+  it('redirects to the licence summary page if the user has not visited it yet', async () => {
+    const mockRequest = createMockRequest()
+    try {
+      await getData(mockRequest)
+    } catch (err) {
+      expect(GetDataRedirect).toBeCalledWith(LICENCE_SUMMARY.uri)
+    }
```
jaucourt, line 16: There's a bit of a trap here, which can exist for tests that throw errors. The problem with the way around you've done this is that if the `getData` function doesn't throw, this test will seemingly pass (in fact, the expect doesn't end up being called, as no error is thrown, so it's just skipped over). If you change the start of `getData` as follows:

```
export const getData = async request => {
  const status = await request.cache().helpers.status.getCurrentPermission()

  if (!status[LICENCE_SUMMARY.page]) {
    // throw new GetDataRedirect(LICENCE_SUMMARY.uri)
  }

  if (!status[CONTACT_SUMMARY.page]) {
    // throw new GetDataRedirect(CONTACT_SUMMARY.uri)
  }

  GetDataRedirect(CONTACT_SUMMARY.uri)
  
  // rest of function as it was...
}
```
All tests still 'pass'.

To overcome this, I think we need to do a couple of things. First of all, we need to modify GetDataRedirect to pass the redirectUrl to error as a message:

```
/**
 * This can be thrown by the get data functions to force a redirect.
 * If specific data items necessary to display the page have not been set
 */

export default class GetDataRedirect extends Error {
  constructor (redirectUrl) {
    super(redirectUrl) // change is here
    this.redirectUrl = redirectUrl
  }
}
```
Then, we can test by doing the following:
```
  it('redirects to the licence summary page if the user has not visited it yet', async () => {
    const mockRequest = createMockRequest()
    await expect(() => getData(mockRequest)).rejects.toThrow(LICENCE_SUMMARY.uri)
  })
```
Could I leave you to modify the other test?

PR #792, opened 2021-07-01
Modify fulfilment-job to read obfuscated DOB from CRM instead of gene…

nabeelamir-defra:
…rating it

https://eaflood.atlassian.net/browse/IWTF-2003

Modify fulfilment-job to read obfuscated DOB from CRM instead of generating it

### DIFF
--- a/packages/dynamics-lib/src/client/__tests__/util.spec.js
+++ b/packages/dynamics-lib/src/client/__tests__/util.spec.js
@@ -1,9 +1,38 @@
-import { escapeODataStringValue } from '../util.js'
+import { escapeODataStringValue, generateDobId, getRandomInt } from '../util.js'
 
 describe('util', () => {
   describe('escapeODataStringValue', () => {
     it('escapes all illegal characters used in a string value of an ODATA $filter clause', async () => {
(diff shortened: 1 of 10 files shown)


packages/dynamics-lib/src/client/__tests__/util.spec.js
```diff
@@ -1,9 +1,38 @@
...
+      expect(randomInt).toBeGreaterThanOrEqual(10)
+      expect(randomInt).toBeLessThanOrEqual(99)
+    })
+
+    it('gets a random integer between a negative range', () => {
+      const randomInt = getRandomInt(-10, 10)
+      expect(randomInt).toBeGreaterThanOrEqual(-10)
+      expect(randomInt).toBeLessThanOrEqual(10)
+    })
+  })
```
jaucourt, line 37: This feels like it might be a bit too close to the implementation - if the random integer generation is only used by the code to generated the obfuscated DOB, then it can probably be kept private within the module and we just test the inputs and outputs - date of birth in, obfuscated date of birth out.

jaucourt (changes requested):
Looks good, just one discussion point

jaucourt (approved):

PR #801, opened 2021-07-08
Feat: Save pocl errors to crm

DGajewska:
IWTF-2174

* Creates PoclDataValidationError entity - in dynamics-lib & temp folder
* Updates POCL File status when errors occur to 'Processed with Warnings'
* Adds call to persist POCL data validation errors when statusCode is 422 and record is present
* Adds pocl data validation error status to option set data


packages/dynamics-lib/src/entities/__tests__/pocl-data-validaton-error.entity.spec.js
```diff
@@ -0,0 +1,157 @@
+import { PoclDataValidationError } from '../pocl-data-validation-error.entity.js'
+import { retrieveGlobalOptionSets } from '../..'
+
+let optionSetData
```
nabeelamir-defra, line 4: I think this could be declared inside the describe block

packages/dynamics-lib/src/index.js
```diff
@@ -12,6 +12,7 @@ export * from './entities/transaction-journal.entity.js'
 export * from './entities/transaction-currency.entity.js'
 export * from './entities/pocl-file.entity.js'
 export * from './entities/pocl-staging-exception.entity.js'
+// export * from './entities/pocl-data-validation-error.entity.js'
```
nabeelamir-defra, line 15: can this comment be removed?

DGajewska: Ah, thank you! This should be uncommented

packages/dynamics-lib/src/entities/__tests__/pocl-data-validaton-error.entity.spec.js
```diff
@@ -0,0 +1,157 @@
...
+        label: 'Post Office Sales',
+        description: 'Post Office Sales'
+      }),
+      status: expect.objectContaining({ id: 910400000, label: 'Needs Review', description: 'Needs Review' })
+    }
+
+    expect(exception).toBeInstanceOf(PoclDataValidationError)
+    expect(exception).toMatchObject(expect.objectContaining({ etag: 'W/"56351087"', ...expectedFields }))
+    expect(exception.toJSON()).toMatchObject(expect.objectContaining(expectedFields))
+    expect(JSON.parse(exception.toString())).toMatchObject(expect.objectContaining(expectedFields))
```
jaucourt, line 95: Multiple expect statements should generally be avoided in favour of several tests. So, have a test checking instance type, then a snapshot test (see below about those). In fact, looks like the next three expects could be replaced by a single snapshot test.

That said, if you do multiple expect statements as the data is taking a lot of mocking, then just have an external function that generates the test data, that you can use in each of your tests

packages/dynamics-lib/src/entities/__tests__/pocl-data-validaton-error.entity.spec.js
```diff
@@ -0,0 +1,157 @@
+import { PoclDataValidationError } from '../pocl-data-validation-error.entity.js'
+import { retrieveGlobalOptionSets } from '../..'
+
+let optionSetData
+describe('pocl staging exception entity', () => {
+  beforeAll(async () => {
+    optionSetData = await retrieveGlobalOptionSets().cached()
```
jaucourt, line 7: I'm not a big fan of `let` declarations - it's probably better to get it for each test rather than getting it once and re-using it. Then the test is self contained.

packages/dynamics-lib/src/entities/__tests__/pocl-staging-exception.entity.spec.js
```diff
@@ -1,8 +1,8 @@
 import { PoclStagingException } from '../pocl-staging-exception.entity.js'
 import { Permission, retrieveGlobalOptionSets } from '../..'
 
-let optionSetData
 describe('pocl staging exception entity', () => {
+  let optionSetData
```
jaucourt, line 5: After expanding this out again, looks like you've got multiple expect statements in this test file too that I'd missed in a previous review - please could you amend these to be separate tests; it's fine to have external functions that get you the test data (so a getSamplePOCLStagingException function, for instance).

packages/pocl-job/src/staging/__tests__/create-transactions.spec.js
```diff
@@ -161,6 +161,49 @@ describe('create-transactions', () => {
...
+          payment: {
+            amount: 82,
+            channelId: '21995X',
+            method: 'Stamps',
+            source: 'Post Office Sales',
+            timestamp: '2020-07-01T12:00:00.000Z'
+          }
+        },
+        id: 'SERIAL 2',
+        stage: 'Transaction Creation Failed'
```
jaucourt, line 206: another one that'd be better as a snapshot test

packages/sales-api-service/src/server/routes/__tests__/staging-exceptions.spec.js
```diff
@@ -37,21 +37,65 @@ describe('staging exceptions handler', () => {
...
+          transactionFile: 'string',
+          permissionId: 'string'
+        }
+        createTransactionFileException.mockResolvedValueOnce(transactionFileException)
+        createDataValidationError.mockResolvedValue()
+      })
+
+      it('adds a transaction file exception', async () => {
+        const result = await server.inject({ method: 'POST', url: '/stagingExceptions', payload: { transactionFileException } })
+        expect(createTransactionFileException).toHaveBeenCalledWith(transactionFileException)
```
jaucourt, line 58: The beforeEach for the describe block needs to be clearing out your mocks: `beforeEach(jest.clearAllMocks)`, otherwise this could be a false positive. Also, should really be multiple tests rather than multiple expect statements here. And the last expect should be a snapshot test.

packages/sales-api-service/src/server/routes/staging-exceptions.js
```diff
@@ -1,8 +1,13 @@
 import { createStagingExceptionRequestSchema, createStagingExceptionResponseSchema } from '../../schema/staging-exception.schema.js'
-import { createTransactionFileException, createStagingException } from '../../services/exceptions/exceptions.service.js'
+import { createTransactionFileException, createStagingException, createDataValidationError } from '../../services/exceptions/exceptions.service.js'
 
 const SWAGGER_TAGS = ['api', 'staging-exceptions']
 
+const isDataValidationError = payload => {
+  const { statusCode } = JSON.parse(payload.transactionFileException.description)
```
jaucourt, line 7: Wouldn't it be tidier to do `payload.code` here? https://hapi.dev/api/?v=20.1.4#-responsecodestatuscode

packages/sales-api-service/src/services/exceptions/__tests__/exceptions.service.spec.js
```diff
@@ -110,4 +111,170 @@ describe('payment-journals service', () => {
...
+            method: 'Cash'
+          }
+        },
+        stage: 'Staging',
+        createTransactionError: {
+          statusCode: 422,
+          error: 'Data validation error',
+          message: 'Error'
+        }
+      }
```
jaucourt, line 160: Rather than declaring this here, please could you extract it into a function - getSampleRecord (or some similar name). Then in each test, it can get its own test data, rather than re-using.

Also, each test should do its own action, so 'maps the record to an instance of PoclDataValidationError' would be:
```
it('maps the record to an instance of PoclDataValidationError', () => {
  const record = getSampleRecord()
  const result = await createDataValidationError(record)
  expect(result).toBeInstanceOf(PoclDataValidationError)
})
```

packages/sales-api-service/src/services/exceptions/__tests__/exceptions.service.spec.js
```diff
@@ -110,4 +111,170 @@ describe('payment-journals service', () => {
...
+        })
+      })
+
+      it('status', () => {
+        expect(result.status.label).toBe('Needs Review')
+      })
+
+      it('data source', () => {
+        expect(result.dataSource.label).toBe('Post Office Sales')
+      })
```
jaucourt, line 277: Pretty much all of these could be replaced by a single snapshot test, I think.

packages/sales-api-service/src/services/exceptions/exceptions.service.js
```diff
@@ -1,6 +1,9 @@
 import { persist, StagingException, PoclStagingException } from '@defra-fish/dynamics-lib'
 import db from 'debug'
 import { getGlobalOptionSetValue } from '../reference-data.service.js'
+
+// @IWTF-2174: Remove after testing
+import { PoclDataValidationError } from './temp/index.js'
```
jaucourt, line 6: is this to be removed once Naimulla has tested? Or can it be removed now?

packages/sales-api-service/src/services/exceptions/temp/index.js
```diff
@@ -0,0 +1 @@
+export * from './pocl-data-validation-error.entity.js'
```
jaucourt, line 1: I'm confused why this is here - couldn't you just use './pocl-data-validation-error.entity.js' directly?

jaucourt (changes requested):
Lots here, but a few changes needed, sorry!

packages/sales-api-service/src/services/exceptions/exceptions.service.js
DGajewska: This will be removed before it is merged into the develop branch, but will need to be left in for now so that I can test the next part of the work and the process as a whole before it's merged in.

packages/sales-api-service/src/services/exceptions/temp/index.js
DGajewska: I did that to begin with, but it didn't want to import it into the file I need it in, so I recreated the way that it is exported in the dynamics-lib and it was happy with this approach. 
This is also to be removed after all the work is completed

packages/sales-api-service/src/services/exceptions/__tests__/exceptions.service.spec.js
DGajewska: I have extracted it into a function. The record is created and then the `createDataValidationError` function is called in the `beforeEach` block, so would be called individually for each test

DGajewska: Based on your comment below, I have changed this to a snapshot test

jaucourt (approved):

PR #802, opened 2021-07-08
Update to Payment Confirmation Screen

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2262

Update to Payment Confirmation Screen

### DIFF
--- a/packages/gafl-webapp-service/src/pages/order-complete/order-complete/order-complete.njk
+++ b/packages/gafl-webapp-service/src/pages/order-complete/order-complete/order-complete.njk
@@ -21,71 +21,69 @@
           titleHtml: titleHtml,
           html: 'Your licence number is<br><strong><span id="permission-number">' + data.permission.referenceNumber + '</span></strong><br /><br />Starts from <br />' + data.startTimeStringTitle
         }) }}
-
+        <p class="govuk-body">You can fish from the time and date your licence starts.</p>
         {% if data.isPostalFulfilment %}
-            <p class="govuk-body">
-                You can fish from the time and date your licence starts. You do not need to wait to get your licence card.
-            </p>
-        {% else %}
-            <p class="govuk-body">You can fish from the time and date your licence starts.</p>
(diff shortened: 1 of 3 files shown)


packages/gafl-webapp-service/src/pages/order-complete/order-complete/order-complete.njk
```diff
@@ -21,71 +21,69 @@
...
-            <p class="govuk-body">
-                It’s important you know your licence number. You can take a screenshot or print out your licence information.
-            </p>
+        {% if data.contactMethod === data.howContacted['none'] %}
+            <p class="govuk-body">Make a note of your licence number and take it with you.</p>
+            <p class="govuk-body">You must give your licence number to an enforcement officer if they ask for it.</p>
+        {% else %}
+            <h2 class="govuk-heading-l">Before you go fishing</h2>
+            {% if data.isPostalFulfilment %}
+                <p class="govuk-body">Make a note of your licence number to take with you. You must give your licence number to an enforcement officer if they ask for it.</p>
```
DGajewska, line 50: Is it possible that both of these are true? So the same content will be displayed twice?

DGajewska (approved):

packages/gafl-webapp-service/src/pages/order-complete/order-complete/order-complete.njk
nabeelamir-defra: Hi Dana, which specific part are you talking about? It should be ok. So there is an if then else. Then in the else, there is an if then else.

DGajewska: Is it possible for this to be true:
`data.contactMethod === data.howContacted['none'] && data.isPostalFulfilment`

nabeelamir-defra: Yes it could, but then only the text in `data.contactMethod === data.howContacted['none']` will be shown.

jaucourt (approved):

PR #817, opened 2021-07-23
Feat: Reprocess POCL validation errors

DGajewska:
IWTF-2175

* Creates Sales Api endpoint to retrieve POCL validation errors 
* Creates Sales Api endpoint to update POCL validation error records
* Adds connector code to new SalesApi endpoints
* Adds reprocessing of POCL validation errors
* Calls reprocessing of POCL validation errors in POCL processor


packages/dynamics-lib/src/queries/__tests__/pocl-validation-error.queries.spec.js
```diff
@@ -0,0 +1,23 @@
...
+      const query = findPoclValidationErrors()
+      expect(query.toRetrieveRequest()).toEqual({
+        collection: 'defra_poclvalidationerrors',
+        filter: 'statecode eq 0',
+        select: expect.any(Array)
+      })
+    })
+  })
+  it('filters by status and default filter if status provided', async () => {
+    const status = { id: 'test-status-id', label: 'Ready for Processing' }
```
nabeelamir-defra, line 15: Very minor, but these 2 tests don't need to be async

nabeelamir-defra (approved):

packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      )
+    })
+
+    it('returns a PoclValidationError instance', () => {
+      expect(exception).toBeInstanceOf(PoclValidationError)
+    })
+
+    it('has the expected data', () => {
+      expect(exception).toMatchSnapshot()
+    })
```
jaucourt, line 54: I've deliberated whether to make this comment - and it is just a comment, do with it what you will. I would do this slightly differently. Rather than having a `beforeAll` I'd have a `getSamplePoclValidationError` function, using this to get me an instance of the class, then assert on that. Something like this
```
it('returns a PoclValidationError instance', () => {
  const exception = getSampleValidationError()
  expect(exception).toBeInstanceOf(PoclValidationError)
})

it('has the expected data', () => {
  const exception = getSampleValidationError()
   expect(exception).toMatchSnapshot()
})
```
I say this because I'd have probably then used the `getSampleValidationError` method to do several different validation errors. 
However, I leave this as a comment, because I'm not all that sure that this would add a huge amount, or possibly even any, value

packages/pocl-job/src/__tests__/pocl-processor.spec.js
```diff
@@ -43,29 +45,64 @@ jest.mock('../config.js', () => ({
...
+      expect(stage).toHaveBeenNthCalledWith(2, 'local/2')
+    })
+
+    it('processes POCL validation errors', () => {
+      expect(processPoclValidationErrors).toHaveBeenCalled()
+    })
+
+    it('releases lock on ETL process', async () => {
+      expect(global.lockReleased).toEqual(true)
+    })
```
jaucourt, line 105: nice work, much easier to see where this fails should it ever do so

packages/sales-api-service/src/schema/contact.schema.js
```diff
@@ -44,7 +45,7 @@ const commonContactSchema = {
   postcode: Joi.when(Joi.ref(DATASOURCE_REF), {
     is: Joi.string().valid(...POCL_TRANSACTION_SOURCES),
     then: Joi.alternatives().conditional('country', {
-      is: Joi.string().valid('GB', 'United Kingdom'),
+      is: Joi.string().valid(...UK_COUNTRIES),
```
jaucourt, line 48: I'm slightly nervous of the inclusion of 'United Kingdom', 'England', 'Wales', 'Scotland' and 'Northern Ireland', as these aren't ISO countries or our handful of exceptions (GB-ENG, etc). So are we passing these now - because I'm not sure we should?

packages/pocl-job/src/staging/__tests__/__snapshots__/create-transactions.spec.js.snap
```diff
@@ -19,6 +19,7 @@ Object {
           "firstName": "Daniel",
           "lastName": "Cleaver",
           "mobilePhone": "07763 722080",
+          "postalFulfilment": undefined,
```
jaucourt, line 22: should this be undefined?

packages/sales-api-service/src/services/exceptions/__tests__/pocl-validation-errors.service.spec.js
```diff
@@ -0,0 +1,170 @@
...
+
+  describe('createPoclValidationError', () => {
+    let payload
+    beforeEach(async () => {
+      payload = getPayload()
+      await createPoclValidationError(payload)
+    })
+
+    it('maps the record to an instance of PoclDataValidationError', async () => {
+      const [poclValidationError] = persist.mock.calls[0][0]
```
jaucourt, line 82: probably do `const [[[poclValidationError]]] = persist.mock.calls` or `const poclValidationError = persist.mock.calls[0][0][0]` rather than mixing the two

packages/sales-api-service/src/services/exceptions/__tests__/pocl-validation-errors.service.spec.js
```diff
@@ -0,0 +1,170 @@
...
+      expect(poclValidationError).toBeInstanceOf(PoclValidationError)
+    })
+
+    it('creates the validation record', async () => {
+      const [poclValidationError] = persist.mock.calls[0][0]
```
jaucourt, line 87: same here...

packages/sales-api-service/src/services/exceptions/__tests__/pocl-validation-errors.service.spec.js
```diff
@@ -0,0 +1,170 @@
...
+          expect(poclValidationError.status.label).toBe('Processed')
+        })
+
+        it('the state code is set to 1', async () => {
+          const [poclValidationError] = persist.mock.calls[0][0]
+          expect(poclValidationError.stateCode).toBe(1)
+        })
+
+        it('updates the validation record', async () => {
+          const [poclValidationError] = persist.mock.calls[0][0]
```
jaucourt, line 143: yeah, here too (as well as the two above), plus any others in the file. I'd say go for the first option (`const [[[poclValidationError]]]`) as it's a little more succinct.

jaucourt (changes requested):
Some great work here, just a few minor comments.

packages/dynamics-lib/src/queries/__tests__/pocl-validation-error.queries.spec.js
DGajewska: Good catch, thanks

packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
DGajewska: I would normally use a `beforeEach` block and either have a function which creates the entity or create it directly in the `beforeEach`. I think I copied the structure of the test file from an existing file and forgot to change that bit.
I've changed the `beforeAll` to a `beforeEach` as I think that will give us the result we want with very little effort

packages/pocl-job/src/__tests__/pocl-processor.spec.js
DGajewska: Thank you :)

packages/sales-api-service/src/schema/contact.schema.js
DGajewska: This schema is used for validating pretty much all records that pass through GAFL. I was more worried about leaving them out and causing some unforseen issues and took the lead from the fact that 'United Kingdom' was included, but I guess all that would happen is that their postcode wouldn't be validated, so I'm happy to remove them

packages/pocl-job/src/staging/__tests__/__snapshots__/create-transactions.spec.js.snap
DGajewska: I don't know, it snuck in after the digital licence work was merged in. I'll double check with Nabeel

packages/sales-api-service/src/services/exceptions/__tests__/pocl-validation-errors.service.spec.js
DGajewska: Ah yes, thank you!

packages/pocl-job/src/staging/__tests__/__snapshots__/create-transactions.spec.js.snap
DGajewska: After a chat with Nabeel (all is okay) and digging into the test data, this was due to the way that the mock for getting permit data was set up. I have updated it to return the correct value and now the flag should be set to `true`

jaucourt (approved):

PR #834, opened 2021-08-02
Removal of Beta badge on GAFL service

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2294

Removal of Beta badge on GAFL service


nabeelamir-defra:
I have removed the use of the govukPhaseBanner component as there is an issue with it. I have opted to use html instead. See https://github.com/alphagov/govuk-frontend/issues/1930 for more info.

jaucourt (approved):

PR #840, opened 2021-08-06
Fix: Sonar cloud issues & test coverage

DGajewska:
IWTF-2175

### DIFF
--- a/packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
+++ b/packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
@@ -18,6 +18,7 @@ describe('pocl staging exception entity', () => {
           defra_organisation: 'Fishy Endeavours',
           defra_premises: '14 Howecroft Court',
           defra_street: 'Eastmead Lane',
+          defra_locality: 'Stoke Bishop',
           defra_town: 'Bristol',
           defra_postcode: 'BS9 1HJ',
           defra_country: 'GB',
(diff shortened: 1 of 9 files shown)


packages/sales-api-service/src/schema/staging-exception.schema.js
```diff
@@ -54,6 +54,8 @@ const schemaObject = {
   }).optional()
 }
 
+const transactionDate = dateSchema.description('An ISO8601 compatible date string defining when the transaction was completed')
```
nabeelamir-defra, line 57: I think this might be personal preference, but because it's a const outside of a function, I think this should be snake case in capitals, so TRANSACTION_DATE.

nabeelamir-defra (approved):

packages/sales-api-service/src/schema/staging-exception.schema.js
DGajewska: Ah yes, good point!

packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
```diff
@@ -52,6 +53,41 @@ describe('pocl staging exception entity', () => {
...
+        paymentSource: 'Post Office Sales',
+        channelId: '948594',
+        methodOfPayment: expect.objectContaining({ id: 910400001, label: 'Cash', description: 'Cash' }),
+        status: expect.objectContaining({ id: 910400000, label: 'Needs Review', description: 'Needs Review' }),
+        dataSource: expect.objectContaining({ id: 910400000, label: 'Post Office Sales', description: 'Post Office Sales' }),
+        stateCode: 1,
+        errorMessage: '\"permissions[0].licensee.email\" must be a valid email'
+      }
+
+      expect(exception).toMatchObject(expect.objectContaining({ etag: 'W/"56351087"', ...expectedFields }))
```
jaucourt, line 89: I'm a bit confused how the snapshot doesn't cover this...?

jaucourt (changes requested):
I think you can dispose of the second test I've commented on as it doesn't do anything the snapshot test doesn't do, otherwise looks good

packages/dynamics-lib/src/entities/__tests__/pocl-validaton-error.entity.spec.js
DGajewska: It does, but SonarCloud doesn't seem to count the snapshot as part of its coverage, so it keeps flagging this file as not having enough test coverage without this added test

jaucourt (approved):

PR #856, opened 2021-08-13
When logging in with one permission and going back and entering anoth…

nabeelamir-defra:
…er permission, the old concession data is cached

https://eaflood.atlassian.net/browse/IWTF-2350

When logging in with one permission and going back and entering another permission, the old concession data is cached

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/renewals-write-cache.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/renewals-write-cache.spec.js
@@ -94,7 +94,7 @@ describe('renewals-write-cache', () => {
     })
 
     it('should set start and end dates, if renewal has not expired', async () => {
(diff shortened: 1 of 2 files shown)


nabeelamir-defra:
I have not set the permission to an empty object as previously discussed. I did this and it did not fix the issue. This is because of the way setCurrentPermission works https://github.com/DEFRA/rod-licensing/blob/b2c833180ac24a3da11b1b86ebb52d6f14c95289/packages/gafl-webapp-service/src/session-cache/cache-decorator.js#L31. It does `Object.assign(current, data)`, so if `current` has the old concession data, and data does not have the concession field, then it won't overwrite it. So the fix is to set the concessions to an empty array.

DGajewska (approved):

jaucourt (approved):

PR #860, opened 2021-08-19
Content changes: display age concession

DGajewska:
IWTF-2362

* Updates page title and url to 'licence details' to match prototype
* Adds age concession to licence details page
* Adds concession helper to get age concession, if exists
* Increases test coverage


packages/gafl-webapp-service/src/processors/concession-helper.js
```diff
@@ -74,6 +74,11 @@ export const removeDisabled = permission => {
   }
 }
 
+export const getAgeConcession = permission =>
+  permission.concessions
+    ? permission.concessions.find(c => c.type !== CONCESSION.DISABLED)
```
nabeelamir-defra, line 79: Very minor and you can leave this as is. But it might be better to be more explicit and say if it's CONCESSION.JUNIOR or CONCESSION.SENIOR

nabeelamir-defra (approved):

packages/gafl-webapp-service/src/processors/concession-helper.js
DGajewska: I had that and then changed it. I couldn't make up my mind! But I do agree with you. I'll change it back

packages/gafl-webapp-service/src/pages/order-complete/licence-details/__tests__/licence-details.spec.js
```diff
@@ -0,0 +1,70 @@
...
+              type: CONCESSION_PROOF.none
+            }
+          }]
+        }))
+      }
+    }
+  }))
+})
+
+afterAll(jest.clearAllMocks)
```
jaucourt, line 36: Why're you doing this - all mocks will be cleared anyway? Should it be `beforeEach(jest.clearAllMocks)`?

packages/gafl-webapp-service/src/pages/order-complete/licence-details/__tests__/licence-details.spec.js
```diff
@@ -0,0 +1,70 @@
...
+            type: CONCESSION.SENIOR,
+            proof: {
+              type: CONCESSION_PROOF.none
+            }
+          }]
+        }))
+      }
+    }
+  }))
+})
```
jaucourt, line 34: Please could you move this to the end of the file? It's purely stylistic, but then you're straight into the tests and can look for this should you need to

packages/gafl-webapp-service/src/processors/concession-helper.js
jaucourt: Or, more succinctly, do
```suggestion
    ? permission.concessions.find(c => [CONCESSION.JUNIOR, CONCESSION.SENIOR].includes(c.type))
```

jaucourt (changes requested):
Couple of minor changes, but nice work :)

packages/gafl-webapp-service/src/processors/concession-helper.js
DGajewska: Yes, you're right Phil, that is the best way

packages/gafl-webapp-service/src/pages/order-complete/licence-details/__tests__/licence-details.spec.js
DGajewska: No, you're right. It should be before each

jaucourt (approved):

PR #903, opened 2021-09-20
Unexpected error is displayed when try to renew another license

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2451

Unexpected error is displayed when try to renew another license

### DIFF
--- a/packages/gafl-webapp-service/src/pages/renewals/identify/route.js
+++ b/packages/gafl-webapp-service/src/pages/renewals/identify/route.js
@@ -5,7 +5,7 @@ import { validation } from '@defra-fish/business-rules-lib'
 
 import GetDataRedirect from '../../../handlers/get-data-redirect.js'
 
-const getData = async (request, h) => {
+const getData = async (request) => {
   // If we are supplied a permission number, validate it or throw 400
   const permission = await request.cache().helpers.status.getCurrentPermission()
 
--- a/packages/gafl-webapp-service/src/pages/renewals/identify/route.js
+++ b/packages/gafl-webapp-service/src/pages/renewals/identify/route.js
@@ -14,7 +14,7 @@ const getData = async request => {
       .permissionNumberUniqueComponentValidator(Joi)
       .validate(permission.referenceNumber)
     if (validatePermissionNumber.error) {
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/pages/renewals/identify/route.js
```diff
@@ -13,7 +14,8 @@ const getData = async request => {
       .permissionNumberUniqueComponentValidator(Joi)
       .validate(permission.referenceNumber)
     if (validatePermissionNumber.error) {
-      throw Boom.forbidden('Attempt to access the authentication page with an invalid permission number')
+      await request.cache().helpers.status.setCurrentPermission({ referenceNumber: '' })
```
DGajewska, line 17: Should this be set to `null` rather than an empty string?

DGajewska (approved):

packages/gafl-webapp-service/src/session-cache/__tests__/session-manager.spec.js
```diff
@@ -28,20 +28,17 @@ describe('isStaticResource', () => {
 })
 
 describe('includesRegex', () => {
-  const regexArray = [/^\/buy\/renew\/identify$/, /^\/renew\/[a-zA-Z0-9]{6}$/]
-  it.each(['/buy/renew/identify', '/renew/ABC123', '/renew/123123', '/renew/ABCDEF'])(
+  const regexArray = [/^\/buy\/renew\/identify$/, /^\/renew\/.*$/]
```
jaucourt, line 31: I don't understand why we're broadening the second regex - are the codes changing? Think we might need a conversation about this.

nabeelamir-defra: So the issue was you can navigate to `/renew/invalid`. Then it would show that 403 forbidden page. This has now been fixed so you redirect to `/buy/renew/identify` instead. However, if you clear your cookies and visit `/renew/invalid` you get redirected to `/buy/date-of-birth`. This is because the `/renew/invalid` was not included in the exemption regex. It was only allowing `/renew\/[a-zA-Z0-9]` e.g. `/renew/MG5GH2`. So I had to make the above change.

jaucourt (approved):

PR #954, opened 2021-10-28
Who is this fishing licence for?

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2380

I want the first screen I interact with on the GAFL service to allow me to determine if I am purchasing a licence for myself or someone else


packages/gafl-webapp-service/src/pages/licence-details/licence-for/__tests__/licence-for.spec.js
```diff
@@ -0,0 +1,42 @@
...
+  })
+
+  it('post response someone-else sets isLicenceForYou to false in status cache', async () => {
+    await injectWithCookies('POST', LICENCE_FOR.uri, {
+      'licence-for': 'someone-else'
+    })
+    const { payload } = await injectWithCookies('GET', TEST_STATUS.uri)
+    expect(JSON.parse(payload).permissions[0].isLicenceForYou).toEqual(false)
+  })
+})
```
jaucourt, line 42: I'd really like us to get rid of this style of test. Please can you remove these tests and add normal unit tests for `licence-for/result-function.js`. You can also add a snapshot test for journey-definition.js if that causes coverage to dip below 100%.

jaucourt (changes requested):
Looks good, but we need to switch out some tests as I've mentioned. Thanks!

packages/gafl-webapp-service/src/pages/licence-details/licence-for/__tests__/route.spec.js
```diff
@@ -0,0 +1,21 @@
...
+      const result = validator.validate({ 'licence-for': 'you' })
+      expect(result.error).toBeUndefined()
+    })
+
+    it('should not return an error if licence-for is someone-else', () => {
+      const result = validator.validate({ 'licence-for': 'someone-else' })
+      expect(result.error).toBeUndefined()
+    })
+  })
+})
```
jaucourt, line 21: So you're testing Joi here, which is fine, but you're not testing the pageRoute at all. You need to mock that and test the inputs.

jaucourt (changes requested):

jaucourt (approved):

PR #961, opened 2021-11-04
What is your dob?

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2382

As an Angler, I want to be requested to enter mine or the individuals DOB once I have entered a name, so that I can proceed to purchase the licence


packages/gafl-webapp-service/src/pages/concessions/date-of-birth/__tests__/route.spec.js
```diff
@@ -0,0 +1,38 @@
...
+      expect(result.isLicenceForYou).toBeFalsy()
+    })
+
+    it('should call getPronoun', async () => {
+      mockStatusCacheGet.mockImplementationOnce(() => ({ isLicenceForYou: true }))
+      await getData(mockRequest)
+      expect(getPronoun).toHaveBeenCalled()
+    })
+  })
+})
```
jaucourt, line 38: You're not testing the default export here, I think you should be

packages/gafl-webapp-service/src/pages/contact/name/__tests__/route.spec.js
```diff
@@ -14,16 +17,22 @@ describe('name > route', () => {
...
       mockStatusCacheGet.mockImplementationOnce(() => ({ isLicenceForYou: false }))
       const result = await getData(mockRequest)
       expect(result.isLicenceForYou).toBeFalsy()
     })
+
+    it('should call getPronoun', async () => {
+      mockStatusCacheGet.mockImplementationOnce(() => ({ isLicenceForYou: true }))
+      await getData(mockRequest)
+      expect(getPronoun).toHaveBeenCalled()
+    })
```
jaucourt, line 36: this test isn't quite right. We need to test the new return value from the getData method (pronoun) is what we'd expect given different scenarios. Whether it calls getPronoun should be incidental - it doesn't matter _how_ it produces its outputs, just that it produces the correct ones.

jaucourt (changes requested):

jaucourt (approved):

PR #969, opened 2021-11-12
Validate licence start time

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-2543

Validate the licence start time more robustly, so that it's not possible
to specify a start time that's out of range (less than the current time
+ 30 minutes) when the start date is today


packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/route.js
```diff
@@ -6,33 +6,55 @@ import moment from 'moment-timezone'
...
+  return { startDateStr, minHour: getMinHour(permission) }
+}
+
+const route = pageRoute(LICENCE_START_TIME.page, LICENCE_START_TIME.uri, validator, nextPage, getData)
+route.find(r => r.method === 'POST').options.ext = {
+  onPostAuth: {
+    method: async (request, reply) => {
+      const permission = await request.cache().helpers.transaction.getCurrentPermission()
+      request.app.permission = permission
+      return reply.continue
```
nabeelamir-defra, line 55: How come this is like this? Should this be in the update-transaction.js?

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/route.js
```diff
@@ -6,33 +6,55 @@ import moment from 'moment-timezone'
...
+    const cantStartUntilTomorrow = moment(now).add(90, 'minute').isAfter(now, 'day')
+    if (cantStartUntilTomorrow) {
+      return 24
+    }
+    return now.add(90, 'minute').startOf('hour').hour()
+  }
+  return 0
+}
+
+const hours = Array(25)
```
nabeelamir-defra, line 24: How come this has changed to 25?

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/route.js
jaucourt: No, because it's not actually modifying the permission, just making it available on the app object. In the validator, this is available whereas the `cache()` function isn't. The validator needs access to the permission as it needs the licence start date to compare to the current date.

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/route.js
jaucourt: Yeah, we need to handle 10:31pm onwards on a day, you'll have seen the `cantStartUntilTomorrow` in minHour, which is for when we've passed the cut off for the day. The minHour function will then return '24', which isn't an available option, so there is no valid option.

PR #974, opened 2021-11-16
Check the licence details

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2577

As a Angler, I want to be able to view the licence details, so that I can verify the details and change any of the licence details if I have made an error

### DIFF
--- a/packages/gafl-webapp-service/assets/sass/rod-licensing/_table_extras.scss
+++ b/packages/gafl-webapp-service/assets/sass/rod-licensing/_table_extras.scss
@@ -1,0 +1,5 @@
+.govuk-summary-list__row--last {
+  dd, dt, span {
+    border-bottom: none;
+  }
+}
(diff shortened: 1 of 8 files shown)


packages/gafl-webapp-service/assets/sass/rod-licensing/_table_extras.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-summary-list__row--last {
+  dd,
+  dt,
+  span {
+    border-bottom: none;
+  }
+}
```
jaucourt, line 7: Not 100% sure what you're doing here with the styling, but rather than applying a class to the last element, you could use the `:last-child` pseudo element selector and then you don't need to add a class. Have a look at what I've done here: https://codepen.io/jaucourt/pen/xxrJKEN

PR #1021, opened 2022-01-05
Update content for BOBO journey screens that are already developed

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2625

Update content for BOBO journey screens that are already developed


jaucourt (changes requested):
Just one wording change, otherwise looks good!

jaucourt (approved):

PR #1031, opened 2022-01-21
Feature/iwtf 2590 attribution redirect

MickStein:
Created pull request for attribution renewal redirect

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
@@ -1,5 +1,6 @@
-import { UTM, ATTRIBUTION_REDIRECT_DEFAULT } from '../../constants'
+import { UTM, ATTRIBUTION_REDIRECT_DEFAULT, QUERYSTRING_LICENCE_KEY } from '../../constants'
 import attributionHandler from '../attribution-handler'
+import { RENEWAL_BASE } from '../../uri'
 
 jest.mock('../../constants', () => ({
   UTM: {
@@ -9,7 +10,14 @@ jest.mock('../../constants', () => ({
     SOURCE: 'utmsource',
     TERM: 'utmterm'
   },
-  ATTRIBUTION_REDIRECT_DEFAULT: '/attribution/redirect/default'
+  ATTRIBUTION_REDIRECT_DEFAULT: '/attribution/redirect/default',
+  reference: {
+    reference: 'reference'
+  }
+}))
+
+jest.mock('../../uri', () => ({
+  RENEWAL_BASE: '/renew'
 }))
 
 describe('The attribution handler', () => {
@@ -45,10 +53,14 @@ describe('The attribution handler', () => {
     })
   })
 
-  it("redirects to ATTRIBUTION_REDIRECT_DEFAULT if ATTRIBUTION_REDIRECT env var isn't set", async () => {
+  it.each([
+    ['campaign-12'],
+    ['sample-campaign'],
+    ['important-campaign']
+  ])("redirects to ATTRIBUTION_REDIRECT_DEFAULT if ATTRIBUTION_REDIRECT env var isn't set", async campaign => {
     delete process.env.ATTRIBUTION_REDIRECT
     const query = {
-      [UTM.CAMPAIGN]: 'campaign-12',
+      [UTM.CAMPAIGN]: campaign,
       [UTM.MEDIUM]: 'click_bait',
       [UTM.CONTENT]: 'eieioh',
       [UTM.SOURCE]: 'tomato',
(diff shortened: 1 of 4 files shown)


nabeelamir-defra (approved):

nabeelamir-defra (changes requested):
Tests appear to be failing

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
```diff
@@ -9,7 +10,14 @@ jest.mock('../../constants', () => ({
     SOURCE: 'utmsource',
     TERM: 'utmterm'
   },
-  ATTRIBUTION_REDIRECT_DEFAULT: '/attribution/redirect/default'
+  ATTRIBUTION_REDIRECT_DEFAULT: '/attribution/redirect/default',
+  reference: {
+    reference: 'reference'
+  }
+}))
```
jaucourt, line 17: reference isn't used, so should be deleted

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
```diff
@@ -9,7 +10,14 @@ jest.mock('../../constants', () => ({
...
+
+jest.mock('../../uri', () => ({
+  RENEWAL_BASE: '/renew'
```
jaucourt, line 20: This needs to change to something other than '/renew', so that we check we don't have anything hard-coded to the value rather than using the RENEWAL_BASE constant. Make it '/licence-renew-url' or something like that.

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
```diff
@@ -75,6 +87,54 @@ describe('The attribution handler', () => {
...
+  it('redirects begins with { RENEWAL_BASE } when journey is renewal', async () => {
+    const query = {
+      [UTM.CAMPAIGN]: 'renewals',
+      [UTM.MEDIUM]: 'click_bait',
+      [UTM.CONTENT]: 'eieioh',
+      [UTM.SOURCE]: 'tomato',
+      [UTM.TERM]: 'Michaelmas'
+    }
+    const responseToolkit = generateResponseToolkitMock()
+    const regExMatch = new RegExp(`${RENEWAL_BASE}`)
```
jaucourt, line 113: This RegExp needs to be anchored to the start of the string, otherwise it'll just be checking that RENEWAL_BASE occurs somewhere in the url (which might give us a passing test, but ultimately not testing what we need it to).
Change it to 
```suggestion
    const regExMatch = new RegExp(`^${RENEWAL_BASE}`)
```

packages/gafl-webapp-service/src/handlers/__tests__/attribution-handler.spec.js
```diff
@@ -75,6 +87,54 @@ describe('The attribution handler', () => {
...
+      [UTM.CAMPAIGN]: 'renewals',
+      [UTM.MEDIUM]: 'click_bait',
+      [UTM.CONTENT]: 'eieioh',
+      [UTM.SOURCE]: 'tomato',
+      [UTM.TERM]: 'Michaelmas',
+      [QUERYSTRING_LICENCE_KEY]: licenceKey
+    }
+    const responseToolkit = generateResponseToolkitMock()
+    await attributionHandler(generateRequestMock(query), responseToolkit)
+    const regExMatch = new RegExp(`^\\${RENEWAL_BASE}\\/${licenceKey}$`)
```
jaucourt, line 134: This looks wrong, I think it should be 
```suggestion
    const regExMatch = new RegExp(`^${RENEWAL_BASE}/${licenceKey}$`)
```

packages/gafl-webapp-service/src/handlers/attribution-handler.js
```diff
@@ -9,5 +10,12 @@ import { initialiseAnalyticsSessionData } from '../processors/analytics.js'
  */
 export default async (request, h) => {
   await initialiseAnalyticsSessionData(request)
+
+  if (request.query[UTM.CAMPAIGN] === 'renewals') {
```
jaucourt, line 14: I think that we need to put the 'renewals' campaign key into a constant, call it `RENEWALS_CAMPAIGN_ID` and put it in constants.js

packages/gafl-webapp-service/src/handlers/attribution-handler.js
```diff
@@ -9,5 +10,12 @@ import { initialiseAnalyticsSessionData } from '../processors/analytics.js'
...
+    if (request.query[QUERYSTRING_LICENCE_KEY]) {
+      return h.redirect(`${RENEWAL_BASE}/${request.query[QUERYSTRING_LICENCE_KEY]}`)
+    }
+    return h.redirect(`${RENEWAL_BASE}`)
```
jaucourt, line 18: Remembering the work we did the other day, I think this might be better going to `IDENTIFY` RATHER THAN `RENEWAL_BASE`. The test would need to be updated as well.

jaucourt (changes requested):

nabeelamir-defra (approved):

jaucourt (approved):

PR #1045, opened 2022-01-26
Alter env var structure

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-2663?atlOrigin=eyJpIjoiYzE1NDQ1NDc4Nzc3NDdjMTljNDg4YzZjMGI0NjYzNjUiLCJwIjoiaiJ9

Change how hold env variables so added to gitignore so not longer part of git repo


nabeelamir-defra (changes requested):
It doesn't say in the story, but I think it's important to do. The README in the docker file should be updated the reflect these changes

jaucourt (changes requested):
This looks good, but it does need the instructions in README.env to be amended

ScottDormand96:
> This looks good, but it does need the instructions in README.env to be amended

I made changes to the README within the docker folder

docker/README.md
```diff
@@ -82,8 +82,24 @@ To support running the services locally using docker, there are three different
...
+git mv gafl-webapp_telesales.secrets.env.example .gafl-webapp_telesales.secrets.env
+git mv gafl-webapp.env.example .gafl-webapp.env
+git mv gafl-webapp.secrets.env.example .gafl-webapp.secrets.env
+git mv payment-mop_up_job.env.example .payment-mop_up_job.env
+git mv payment-mop_up_job.secrets.env.example .payment-mop_up_job.secrets.env
+git mv pocl_job.env.example .pocl_job.env
+git mv sales_api.env.example .sales_api.env
+git mv sales_api.secrets.env.example .sales_api.secrets.env
+git mv sqs_receiver.env.example .sqs_receiver.env
+```
```
jaucourt, line 102: You don't want to do `git mv` as that'd rename them within git. `cp fulfilment_job.env.example .fulfilment_job.env`, etc, would do what's needed as we want to keep the original

docker/README.md
```diff
@@ -82,8 +82,24 @@ To support running the services locally using docker, there are three different
   > Building the containers in production mode takes longer as all node modules need to be installed and additional build
   > steps such as compiling SASS need to be performed.
 
-In order to run the services locally, you'll need to insert the appropriate values into the environment files ending with .secrets.env
-in the [env](env) folder.
+In order to run the services locally, you'll need to to rename the env files in the in the [env](env) folder to include a leading dot and removing .example. You'll need to insert the appropriate values into the environment files ending with .secrets.env. Either rename as below or grab the filles from the the gitlab repository: fish/rod-licensing-env-vars
```
jaucourt, line 85: This isn't quite correct: 'Either rename as below or grab the filles from the the gitlab repository: fish/rod-licensing-env-vars' should be something like 'Copy files as follows, then get values for secret files from gitlab repo fish/rod-licensing-env-vars.'

.gitignore
```diff
@@ -37,4 +37,7 @@ docker/volumes/**/*
 # Ignore any hidden files/folders (starting with a .)
 **/.*
 # except the github dir...
-!.github 
\ No newline at end of file
+!.github 
+
+# Ignore any docker env files
+*.env
```
jaucourt, line 43: Thinking about it, we won't need this as we're already ignoring files starting with a dot, above

jaucourt (changes requested):
Sorry, few more changes needed

jaucourt (approved):

nabeelamir-defra (approved):

PR #1074, opened 2022-02-11
Easy renewals vanity URL

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-2720?atlOrigin=eyJpIjoiMGQ1ZGVhNmZkYjE0NDdiODhlN2UxMTc3MmQzN2UyODAiLCJwIjoiaiJ9

updating to a friendly 'vanity' url for easy renewals attribution

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/url-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/url-handler.spec.js
@@ -1,6 +1,6 @@
 import { UTM, QUERYSTRING_LICENCE_KEY } from '../../constants'
 import urlHandler from '../url-handler'
-import { RENEWAL_LICENCE, IDENTIFY } from '../../uri'
+import { ATTRIBUTION, IDENTIFY } from '../../uri'
 
 jest.mock('../../constants', () => ({
   UTM: {
@@ -13,7 +13,7 @@ jest.mock('../../constants', () => ({
 }))
 
 jest.mock('../../uri', () => ({
-  RENEWAL_LICENCE: { uri: '/licence-renew-url' },
+  ATTRIBUTION: { uri: '/attribution-url' },
   IDENTIFY: { uri: '/renewal-url' }
 }))
 
@@ -22,7 +22,7 @@ describe('The url handler', () => {
     jest.clearAllMocks()
   })
 
-  it('test returns identify if no licence key', async () => {
+  it('returns identify if no licence key', async () => {
     const query = {
       [UTM.CAMPAIGN]: 'renewals',
       [QUERYSTRING_LICENCE_KEY]: null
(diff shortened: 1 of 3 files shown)


packages/gafl-webapp-service/src/uri.js
```diff
@@ -48,6 +48,7 @@ export const RENEWAL_INACTIVE = { uri: '/buy/renew/inactive', page: 'renewal-ina
 export const AUTHENTICATE = { uri: '/buy/renew/authenticate' }
 export const RENEWAL_START_DATE = { uri: '/buy/renew/renewal-start-date', page: 'renewal-start-date' }
 export const RENEWAL_START_VALIDATE = { uri: '/buy/renew/renewal-start-date/validate' }
+export const RENEWAL_LICENCE = { uri: '/renew-my-licence/{reference}', page: '/attribution?utm_campaign=renewals&urm_source=aen_invitation&reference=<licence number>' }
```
jaucourt, line 51: page shouldn't be defined like this - if you look, no other uri constant is set up like that. Also, unless I've missed something, you don't actually use the page property. I think you meant to, see my comment on your test, but it should either be setup in the journey definition or by using `ATTRIBUTION` above and building a query string.

packages/gafl-webapp-service/src/handlers/__tests__/url-handler.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    ['GH330P']
+  ])('test 6 digit reference number exists and returns RENEWAL_LICENCE', async licenceKey => {
+    const query = {
+      [UTM.CAMPAIGN]: 'renewals',
+      [QUERYSTRING_LICENCE_KEY]: licenceKey
+    }
+    const responseToolkit = generateResponseToolkitMock()
+    await urlHandler(generateRequestMock(query), responseToolkit)
+    const regExMatch = new RegExp(`^${RENEWAL_LICENCE.uri}/${licenceKey}$`)
+    expect(responseToolkit.redirect).toHaveBeenCalledWith(expect.stringMatching(regExMatch))
```
jaucourt, line 47: surely it should be redirecting to the attribution page, not itself?

packages/gafl-webapp-service/src/handlers/url-handler.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default async (request, h) => {
+  await initialiseAnalyticsSessionData(request)
+  if (request.query[QUERYSTRING_LICENCE_KEY]) {
+    const refNumber = request.query[QUERYSTRING_LICENCE_KEY]
+    const sixDigit = /^[A-Za-z0-9]{6}$/.test(refNumber)
+    if (!sixDigit) {
+      return h.redirect(IDENTIFY.uri)
+    }
+    return h.redirect(`${RENEWAL_LICENCE.uri}/${refNumber}`)
```
jaucourt, line 19: the above could be simplified to:
```
if (sixDigit) {
  return h.redirect(`${RENEWAL_LICENCE.uri}/${refNumber}`)
}
```
if `sixDigit` is false, it'll fall through to the redirect below

jaucourt (changes requested):

packages/gafl-webapp-service/src/uri.js
ScottDormand96: ready for re review

packages/gafl-webapp-service/src/handlers/__tests__/url-handler.spec.js
ScottDormand96: ready for re review

jaucourt (approved):

PR #1107, opened 2022-02-23
Fix Sonar Cloud QA failure

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2767

Issues are in renewals-write-cache, with params that are not used and should be removed.


nabeelamir-defra (approved):

packages/gafl-webapp-service/src/processors/renewals-write-cache.js
```diff
@@ -33,14 +33,16 @@ export const setUpCacheFromAuthenticationResult = async (request, authentication
...
-    ),
-    {
-      countryCode: authenticationResult.permission.licensee.country.description
-    }
-  )
+  const copyLicenseeWithoutContactAndCountry = ({
+    country,
+    _preferredMethodOfConfirmation,
+    _preferredMethodOfNewsletter,
+    _preferredMethodOfReminder,
```
nabeelamir-defra, line 40: May need Phil's input on this, will it work with the _ prefixed?

nabeelamir-defra: It would be worth adding a test to say that the 4 are omitted from `permission.licensee`

jaucourt (changes requested):

PR #1133, opened 2022-03-07
Update address lookup capital formatting

MickStein:


packages/gafl-webapp-service/src/pages/contact/address/select/address-select.njk
```diff
@@ -14,7 +14,7 @@
 
 {% for address in data.addresses %}
     {{ addressItems.push({
-        text: address.address,
+        text: address.address_full,
```
jaucourt, line 17: here, you'd simply access the address, so leave it unchanged

packages/gafl-webapp-service/src/pages/contact/address/select/address-select.njk
```diff
@@ -32,6 +32,7 @@
 
     {{ govukRadios({
         idPrefix: "address",
+        classes: "capitalise",
```
jaucourt, line 35: Semantic class names should be preferred, so use 'address_list' rather than 'capitalise'

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -44,7 +46,25 @@ export default async (premises, postcode) => {
       locality: r.locality,
       town: r.city,
       postcode: r.postcode,
-      country: r.country
+      country: r.country,
+      address_full: formatAddress(r.premises, r.street_address, r.locality, r.city, r.postcode)
```
jaucourt, line 50: rather than doing this, modify the address property to be:
```
address: `${r.address.replace(r.postcode, '').toLowerCase()}${r.postcode}`
```
transforming the address, apart from the postcode, to be all lowercase allows you to use the css property text-transform, as the capitalise value will transform lower-case text, but not upper case text

jaucourt (changes requested):
I'm so sorry, because you've obviously put a significant amount of work into this, but I think it could be done more simply by using css functionality. I've put comments in the relevant places of the PR, but ideally we'd avoid our own capitalise function that we'd then have to maintain and would add to the overall code complexity.

In addition to my suggestions, you'd need a new scss file (call it _address.scss) in the assets/sass/rod_licensing directory, and then added into main.scss. To get main.css to compile with your changes, do `npm run prepare` within the packages/gafl-webapp-service directory

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -3,15 +3,46 @@ import fetch from 'node-fetch'
...
+          id: 0,
+          address: '1 howecroft court, eastmead lane, bristol, BS9 1HJ',
+          premises: '1',
+          street: 'HOWECROFT COURT',
+          locality: 'EASTMEAD LANE',
+          town: 'BRISTOL',
+          postcode: 'BS9 1HJ',
+          country: undefined
+        }
+      ])
```
jaucourt, line 45: I think we're testing stuff here that we're not interested in. The only properties that actually get used are address and postcode, so I'd just put those into the mock data. Then, it'll be a simpler process to have multiple cases - I'd put three in so there's a bit of variation in the data.

jaucourt (changes requested):
Looks great, just one last change and I think we're there. Thanks!

jaucourt (approved):

PR #1151, opened 2022-03-17
Welsh Language: Implement feature switch for Welsh language toggle

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2797

Welsh Language: Implement feature switch for Welsh language toggle


jaucourt (approved):
I think this will need an update to https://gitlab-dev.aws-int.defra.cloud/fish/rod-licensing-env-vars as well as new variables setup in AWS. The release notes for the new version of GAFL should also be updated with these details so that we can update Webops (I've just added a new page here: https://eaflood.atlassian.net/wiki/spaces/IWTF/pages/3905847366/.+.22+GAFL+-+P+standard+change).

PR #1166, opened 2022-03-28
Remove aws-sdk dependency from main package.json

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2746

Remove aws-sdk dependency from main package.json


packages/fulfilment-job/src/__mocks__/aws-sdk.js
```diff
@@ -1,3 +1,3 @@
 import { configureAwsSdkMock } from '../../../connectors-lib/src/__mocks__/aws-mock-helper.js'
-const AwsSdk = configureAwsSdkMock(jest.genMockFromModule('aws-sdk'))
+const AwsSdk = configureAwsSdkMock()
```
jaucourt, line 2: Could we change the import to '@defra-fish/connectors-lib/src/__mocks__/aws-mock-helper.js' rather than what it is?

jaucourt (changes requested):

jaucourt (approved):

PR #1213, opened 2022-04-11
Welsh Translation - How long want licence for

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-2708

Update How long do you want your licence for? page to include welsh translation


packages/gafl-webapp-service/src/pages/macros/licence-type-summary.njk
```diff
@@ -0,0 +1,19 @@
...
+        {% elseif licenceTypeStr === 'Over 65, Trout and coarse, up to 3 rods' %}
+            {{ mssgs.over_65 + mssgs.licence_type_radio_trout_three_rod }}
+        {% elseif licenceTypeStr === 'Over 65, Salmon and sea trout' %}
+            {{ mssgs.over_65 + mssgs.licence_type_radio_salmon }}
+        {% elseif licenceTypeStr === 'Trout and coarse, up to 2 rods' %}
+            {{ mssgs.licence_type_radio_trout_two_rod }}
+        {% elseif licenceTypeStr === 'Trout and coarse, up to 3 rods' %}
+            {{ mssgs.licence_type_radio_trout_three_rod }}
+        {% elseif licenceTypeStr === 'Salmon and sea trout' %}
+            {{ mssgs.licence_type_radio_salmon }}
```
nabeelamir-defra, line 14: This should really be a translation of https://github.com/DEFRA/rod-licensing/blob/4eb1cddb3bb1bd9b0cb0bd4d27a633953eb59a1b/packages/gafl-webapp-service/src/processors/licence-type-display.js#L4 and then the js method licenceTypeDisplay would not be called in the route.js. Otherwise the js method licenceTypeDisplay is getting called and this njk licenceTypeSummary method. It adds complexity

nabeelamir-defra: You could do something like, in route.js
```
return {
 isSenior: concessionHelper.hasSenior(permission),
 permission,
 mappings
}
```

Then in njk, (code is untested)
```
typesStrArr= []
if(isSenior) {
 typesStrArr.push(mssgs.over_65)
}
typesStrArr.push(mssgs[permission.licenceType])
if (permission.licenceType === mappings.LICENCE_TYPE['trout-and-coarse']) {
    typesStrArr.push(permission.numberOfRods === '2' ? mssgs.rod_2 : mssgs.rod_2)
  }

return typesStrArr.join(', ')
```
Then you won't have to call licenceTypeDisplay in the route.js

jaucourt: I agree with Nabeel, but I'd go further and say that you should construct the messages you need in `getData`, so that you don't have any of this logic in a nunjucks template. The problem with having logic in a nunjucks template is that it's untested, whereas we can have unit tests for the getData function

jaucourt (changes requested):

nabeelamir-defra (approved):

jaucourt (approved):

PR #1268, opened 2022-05-09
Welsh translations BOBO Manual address entry

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2828

Update welsh translation for manual address entry page.


MickStein:
The ticket is for bobo and should only have a title update. The Jira ticket will need to be reviewed further, I will close this PR and start a new one.

PR #1279, opened 2022-05-11
Update templating in display-licence-type processor and add tests

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2900

The licence summary screen type result currently is not templated, as it is processed in the display-licence-type processor. This needs to be updated and tests adde.


MickStein:
This is being dealt with in a similar PR by Scott, due to this I will close.

PR #1280, opened 2022-05-11
Is the phone number correct - welsh translation

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2713

Is the phone number correct - welsh translation


jaucourt (changes requested):
I think this is missing some of the ACs

ScottDormand96 (approved):

jaucourt (approved):

PR #1329, opened 2022-06-06
Open OS terms footer link in a new tab

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2850

This link should open in a new tab, and say that it opens in a new tab.

### DIFF
--- a/packages/gafl-webapp-service/src/locales/en.json
+++ b/packages/gafl-webapp-service/src/locales/en.json
@@ -79,7 +79,7 @@
   "address_lookup_content_info": "This is where we will send the licence card. A bankside enforcement officer may ask for this information if the licence is checked.",
   "address_lookup_crown_copyright_new_tab": "OS (opens in a new tab)",
   "address_lookup_crown_copyright_ref": "100024198.",
-  "address_lookup_crown_copyright_terms_link": "terms and conditions",
+  "address_lookup_crown_copyright_terms_link": "terms and conditions (opens in a new tab)",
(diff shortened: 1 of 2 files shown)


MickStein (approved):

packages/gafl-webapp-service/src/locales/en.json
```diff
@@ -79,7 +79,7 @@
   "address_lookup_content_info": "This is where we will send the licence card. A bankside enforcement officer may ask for this information if the licence is checked.",
   "address_lookup_crown_copyright_new_tab": "OS (opens in a new tab)",
   "address_lookup_crown_copyright_ref": "100024198.",
-  "address_lookup_crown_copyright_terms_link": "terms and conditions",
```
jaucourt, line 82: I think we'll need this in Welsh too. I've asked Joe and Melinda to confirm what they want, but I think the Welsh translation of "telerau a'r amodau" should be amended to "telerau a'r amodau (yn agor ar dudalen newydd)"

jaucourt (commented):
Need to check with Joe & Melinda on the Welsh translation, otherwise this is fine

packages/gafl-webapp-service/src/locales/en.json
irisfaraway: @jaucourt Since we've agreed to handle the translation in a separate ticket, is it OK to merge this PR now?

jaucourt: Yes - approved

PR #1351, opened 2022-06-16
Welsh Translation for "Free" in the GAFL Pages

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2926

Update welsh translations for where free is shown in the journey including the pay display macro when buying a licence.


MickStein:
This has been updated in a previous commit

PR #1367, opened 2022-06-28
Identify multibuy licences

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-2928

Add a flag to identify multibuy licences

### DIFF
--- a/packages/dynamics-lib/src/entities/__tests__/permission.entity.spec.js
+++ b/packages/dynamics-lib/src/entities/__tests__/permission.entity.spec.js
@@ -6,8 +6,8 @@ describe('permission entity', () => {
     optionSetData = await retrieveGlobalOptionSets().cached()
   })
 
-  it('maps from dynamics', async () => {
-    const permission = Permission.fromResponse(
+  describe('maps from dynamics', () => {
+    const getPermission = () => Permission.fromResponse(
       {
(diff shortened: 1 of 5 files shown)


packages/dynamics-lib/src/entities/__tests__/permission.entity.spec.js
```diff
@@ -6,37 +6,53 @@ describe('permission entity', () => {
...
       referenceNumber: '00000000-2WC3FDR-CD379B',
       issueDate: '2019-12-13T09:00:00Z',
       startDate: '2019-12-14T00:00:00Z',
       endDate: '2020-12-13T23:59:59Z',
       stagingId: '71ad9a25-2a03-406b-a0e3-f4ff37799374',
       dataSource: expect.objectContaining({ id: 910400003, label: 'Web Sales', description: 'Web Sales' }),
-      isLicenceForYou: expect.objectContaining({ id: 1, label: 'Yes', description: 'Yes' })
-    }
+      isLicenceForYou: expect.objectContaining({ id: 1, label: 'Yes', description: 'Yes' }),
+      isMultiBuy: true
```
ScottDormand96, line 36: Looks good, only comment being do we want to check that state is not set when isn't a multi buy?

PR #1376, opened 2022-07-05
Feature/iwtf 2934 welsh month and days

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-2934

Update moment to update based on language

### DIFF
--- a/packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
+++ b/packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
@@ -1,6 +1,7 @@
 import route from '../route.js'
 import moment from 'moment-timezone'
 import pageRoute from '../../../../routes/page-route.js'
+import { cacheDateFormat } from '../../../../processors/date-and-time-display.js'
 
 jest.mock('../../../../routes/page-route.js', () =>
   jest.fn(() => [
@@ -10,6 +11,11 @@ jest.mock('../../../../routes/page-route.js', () =>
     }
   ])
 )
+
+jest.mock('../../../../processors/date-and-time-display.js', () => ({
+  cacheDateFormat: 'YYYY-MM-DD'
+}))
+
 jest.mock('moment-timezone')
 const realMoment = jest.requireActual('moment-timezone')
 
@@ -180,18 +186,19 @@ describe('getData', () => {
     expect(startDateStr).toEqual(formatSymbol)
   })
 
-  it.each([['2020-07-06'], ['2020-01-01'], ['2022-07-09']])('permission licence start date is passed to moment', async licenceStartDate => {
+  it('permission licence start date is passed to moment', async () => {
+    const licenceStartDate = Symbol('licenceStartDate')
     mockTransactionCacheGet.mockImplementationOnce(() => ({ licenceStartDate }))
     moment.mockImplementation(getMomentMockImpl())
     await getData(mockRequest)
-    expect(moment).toHaveBeenCalledWith(licenceStartDate, 'YYYY-MM-DD', 'en-gb')
+    expect(moment).toHaveBeenCalledWith(licenceStartDate, expect.any(String), expect.any(String))
   })
 
   it('cacheDateFormat being what is expected', async () => {
     mockTransactionCacheGet.mockImplementationOnce(() => ({ licenceStartDate: '2020-01-01' }))
     moment.mockImplementation(getMomentMockImpl())
     await getData(mockRequest)
-    expect(moment).toHaveBeenCalledWith('2020-01-01', 'YYYY-MM-DD', 'en-gb')
+    expect(moment).toHaveBeenCalledWith(expect.any(String), cacheDateFormat, expect.any(String))
   })
 
   it('dddd, Do MMMM, YYYY is passed to format', async () => {
@@ -202,7 +209,8 @@ describe('getData', () => {
     expect(format).toHaveBeenCalledWith('dddd, Do MMMM, YYYY')
   })
 
-  it.each([['en-gb'], ['cy']])('locale is set on moment, to whatever the request.locale is', async language => {
+  it('locale is set on moment, to whatever the request.locale is', async () => {
+    const language = Symbol('locale')
     const mockRequest = {
       cache: () => ({
         helpers: {
@@ -217,6 +225,6 @@ describe('getData', () => {
     const format = jest.fn()
     moment.mockImplementation(getMomentMockImpl({ format }))
     await getData(mockRequest)
-    expect(moment).toHaveBeenCalledWith('2020-01-01', 'YYYY-MM-DD', language)
+    expect(moment).toHaveBeenCalledWith(expect.any(String), expect.any(String), language)
   })
 })
--- a/packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
+++ b/packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
@@ -1,4 +1,4 @@
-import startTimePageRoute from '../route.js'
+import route from '../route.js'
 import moment from 'moment-timezone'
 import pageRoute from '../../../../routes/page-route.js'
 
@@ -12,6 +12,13 @@ jest.mock('../../../../routes/page-route.js', () =>
 )
 jest.mock('moment-timezone')
 const realMoment = jest.requireActual('moment-timezone')
+
+const getMomentMockImpl = (overrides = {}) =>
+  jest.fn(() => ({
+    tz: () => ({ isSame: () => {} }),
+    format: () => {},
+    ...overrides
+  }))
 
 describe('Licence start time data validation', () => {
   const validator = pageRoute.mock.calls[1][2]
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
```diff
@@ -137,3 +144,79 @@ describe('licenceStartTimeRoute > onPostAuth', () => {
...
+    })
+    expect(startDateStr).toEqual(formatSymbol)
+  })
+
+  it.each([['2020-07-06'], ['2020-01-01'], ['2022-07-09']])('permission licence start date is passed to moment', async licenceStartDate => {
+    mockTransactionCacheGet.mockImplementationOnce(() => ({ licenceStartDate }))
+    moment.mockImplementation(getMomentMockImpl())
+    await getData(mockRequest)
+    expect(moment).toHaveBeenCalledWith(licenceStartDate, 'YYYY-MM-DD', 'en-gb')
+  })
```
jaucourt, line 188: This could be simplified a bit - I think you could make licenceStartDate a `Symbol` and then you wouldn't need different test cases, just one. Also, you don't care what the second and third arguments are, so do:
```
expect(moment).toHaveBeenCalledWith(licenceStartDate, expect.any(String), expect.any(String))
```

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
```diff
@@ -137,3 +144,79 @@ describe('licenceStartTimeRoute > onPostAuth', () => {
...
+
+  it('cacheDateFormat being what is expected', async () => {
+    mockTransactionCacheGet.mockImplementationOnce(() => ({ licenceStartDate: '2020-01-01' }))
+    moment.mockImplementation(getMomentMockImpl())
+    await getData(mockRequest)
+    expect(moment).toHaveBeenCalledWith('2020-01-01', 'YYYY-MM-DD', 'en-gb')
+  })
```
jaucourt, line 195: cacheDateFormat is imported from `date-and-time-display.js`, so I think you need to mock it (again, as a Symbol), then (as you don't care what the first and third arguments are) assert with:
```
expect(moment).toHaveBeenCalledWith(expect.any(String), mockCacheDateFormat, expect.any(String))
```
Then you need a test to check that request.locale is passed to moment - use the same approach as you have for this and the previous test, mocking `request.locale` as a `Symbol` again.

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
```diff
@@ -137,3 +144,79 @@ describe('licenceStartTimeRoute > onPostAuth', () => {
...
+          }
+        }
+      }),
+      locale: language
+    }
+    mockTransactionCacheGet.mockImplementationOnce(() => ({ licenceStartDate: '2020-01-01' }))
+    const format = jest.fn()
+    moment.mockImplementation(getMomentMockImpl({ format }))
+    await getData(mockRequest)
+    expect(moment).toHaveBeenCalledWith('2020-01-01', 'YYYY-MM-DD', language)
```
jaucourt, line 220: Ah, I see you're testing the locale here - good stuff, but as I say you don't care what the first and second arguments are, so use `expect.any(String)`, then use a `Symbol` for locale so you won't need several test cases.

We will need one final test, to check the return value of moment().format() is passed back as `startDateStr`

jaucourt (changes requested):
Good stuff here Scott, just needs a little bit of tweaking

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,225 @@
...
+
+    it('dateDisplayFormat is passed to format', async () => {
+      mockStatusCacheGet.mockImplementationOnce(() => ({
+        authentication: { reason: RENEWAL_ERROR_REASON.NOT_DUE, endDate: '2020-12-13T23:59:59Z' },
+        referenceNumber: 'ABC123'
+      }))
+      const format = jest.fn()
+      moment.mockImplementation(getMomentMockImpl({ format }))
+      await getData(mockRequest)
+      expect(format).toHaveBeenCalledWith('D MMMM YYYY')
```
jaucourt, line 222: I think the same comments apply here - focus in the `expect` statements on a single argument, using `expect.any(type)`.

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/route.spec.js
ScottDormand96: pushed all changes, tests for return value of moment().format() is passed back as startDateStr already exists line 168

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,230 @@
...
+
+    it('dateDisplayFormat is passed to format', async () => {
+      mockStatusCacheGet.mockImplementationOnce(() => ({
+        authentication: { reason: RENEWAL_ERROR_REASON.NOT_DUE, endDate: '2020-12-13T23:59:59Z' },
+        referenceNumber: 'ABC123'
+      }))
+      const format = jest.fn()
+      moment.mockImplementation(getMomentMockImpl({ format }))
+      await getData(mockRequest)
+      expect(format).toHaveBeenCalledWith(dateDisplayFormat)
```
jaucourt, line 227: Again, you're mocking it now, but you're not using a Symbol. Fine for most tests, just override it as a Symbol for this one.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,236 @@
...
         }
-      }
-    }),
-    i18n: {
-      getCatalog: jest.fn(() => englishTranslations)
+      }),
+      i18n: {
+        getCatalog: jest.fn(() => englishTranslations)
+      },
+      locale: 'en-gb'
```
jaucourt, line 110: This is a bit messy - there's already a function to get a mock request, this should be rolled into it. Also, we shouldn't be importing English and Welsh translations, these should be mocked otherwise there's a danger of tight coupling.

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,236 @@
...
+            status: {
+              getCurrentPermission: mockStatusCacheGet
+            }
+          }
+        }),
+        i18n: {
+          getCatalog: jest.fn(() => englishTranslations)
+        },
+        locale: language
+      }
```
jaucourt, line 211: Another case of a mock request being created, should use the getMockRequest function

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,236 @@
...
+      moment.mockImplementation(getMomentMockImpl())
       const result = await getData(mockRequest)
       expect(result.uri.new).toBe(NEW_TRANSACTION.uri)
     })
 
-    it('should return the body message', async () => {
+    it.each([
+      ['2020-12-13T23:59:59Z', RENEWAL_ERROR_REASON.NOT_DUE, 'ABV123'],
+      ['2020-01-13T23:59:59Z', RENEWAL_ERROR_REASON.NOT_DUE, 'LKJ234'],
+      ['2020-10-13T23:59:59Z', RENEWAL_ERROR_REASON.NOT_DUE, 'MHG356']
```
jaucourt, line 141: for every one of these, `RENEWAL_ERROR_REASON.NOT_DUE` is passed as a parameter. Given it's the same every time, it's really not a parameter. Should it vary? If not, pass it as a static value.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/renewals/renewal-inactive/__tests__/route.spec.js
```diff
@@ -1,45 +1,218 @@
...
+const welshTranslations = {
+  renewal_inactive_title_1: 'You are renewing this licence too early',
+  renewal_inactive_title_2: 'The licence renewal has expired',
+  renewal_inactive_title_3: 'You cannot renew an 8 day or 1 day licence',
+  renewal_inactive_not_due_1: 'The licence ending in ',
+  renewal_inactive_not_due_2: ' does not expire until ',
+  renewal_inactive_has_expired_1: ' has expired on ',
+  renewal_inactive_has_expired_2: ' and can no longer be renewed',
+  renewal_inactive_not_annual_1: ' is not a 12 month licence and cannot be renewed.'
+}
```
jaucourt, line 32: These are identical to `englishTranslations`, above, and as far as I can see nothing is asserted against either - so you could get rid of them and change `englishTranslations` to `sampleLabels`, or something like that. Better than that, however, might be moving the getMessages function below to here and using it whenever you need labels, not just for the block it's in.

jaucourt (changes requested):
Thanks for all your effort here Scott. Just one last little thing and I think we're there.

PR #1410, opened 2022-08-01
Pass current language to GOV.UK Pay API

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2897

If the user has chosen to use the Welsh language for our service, it should also be used for GOV.UK Pay too.

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
@@ -118,33 +118,14 @@ describe('preparePayment', () => {
     expect(ret).toEqual(returnValue)
   })
 
-  describe('provides the correct language', () => {
-    it('when the language is set to Welsh', () => {
-      addLanguageCodeToUri.mockReturnValue('https://localhost:1234/buy/agreed?lang=cy')
-
-      const result = preparePayment(request, transaction)
-      const ret = result.language
-
-      expect(ret).toEqual('cy')
-    })
-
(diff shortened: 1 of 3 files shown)


packages/gafl-webapp-service/src/processors/__tests__/payment.spec.js
```diff
@@ -118,6 +118,35 @@ describe('preparePayment', () => {
...
+    })
+
+    it('when the language is not set', () => {
+      addLanguageCodeToUri.mockReturnValue('https://localhost:1234/buy/agreed')
+
+      const result = preparePayment(request, transaction)
+      const ret = result.language
+
+      expect(ret).toEqual('en')
+    })
```
jaucourt, line 147: I think these three could be one `it.each` test:
```
it.each([
  ['when the language is set to Welsh', 'https://localhost:1234/buy/agreed?lang=cy', 'cy'],
  ['when the language is set to English', 'https://localhost:1234/buy/agreed?lang=en', 'en'],
  ['when the language is not set', 'https://localhost:1234/buy/agreed', 'en']
])('provides the correct language %s', (_desc, decoratedUrl, expectedLanguageCode) => {
  addLanguageCodeToUri.mockReturnValue(decoratedUrl)
  const {language} = preparePayment(request, transaction)
  expect(language).toEqual(expectedLanguageCode)
})
```

jaucourt (changes requested):

jaucourt (approved):

PR #1411, opened 2022-08-01
Update end date and time rules for 12-month licences

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2889

12-month licences should expire at 23:59 on the eve of the anniversary of purchase.

For example, if I bought a licence today on 1 August 2022, it would expire at 23:59 on 31 July 2023.

### DIFF
--- a/packages/sales-api-service/src/services/__tests__/permissions.service.spec.js
+++ b/packages/sales-api-service/src/services/__tests__/permissions.service.spec.js
@@ -46,18 +46,12 @@ describe('permissions service', () => {
           licensee: {
             firstName: 'Fester',
             lastName: 'Tester',
-            birthDate: moment(now)
-              .subtract(JUNIOR_MAX_AGE, 'years')
-              .format('YYYY-MM-DD')
+            birthDate: moment(now).subtract(JUNIOR_MAX_AGE, 'years').format('YYYY-MM-DD')
           }
         },
         'Telesales'
       )
-      const block1 = moment(now)
-        .add(1, 'hour')
-        .startOf('hour')
-        .add(1, 'year')
-        .format('HHDDMMYY')
+      const block1 = moment(now).add(1, 'year').subtract(1, 'day').endOf('day').format('HHDDMMYY')
       const expected = new RegExp(`^${block1}-1TS3FFT-[A-Z0-9]{5}[0-9]$`)
       expect(number).toMatch(expected)
     })
@@ -79,11 +73,7 @@ describe('permissions service', () => {
         },
         'Web Sales'
       )
-      const block1 = moment(now)
-        .add(1, 'hour')
-        .startOf('hour')
-        .add(1, 'day')
-        .format('HHDDMMYY')
+      const block1 = moment(now).add(1, 'hour').startOf('hour').add(1, 'day').format('HHDDMMYY')
       const expected = new RegExp(`^${block1}-2WC1JFT-[A-Z0-9]{5}[0-9]$`)
       expect(number).toMatch(expected)
     })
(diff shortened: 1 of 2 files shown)


irisfaraway:
Looks like the automated linter decided to make all those moment calls one-liners – hope that's alright!

packages/sales-api-service/src/services/__tests__/permissions.service.spec.js
```diff
@@ -98,41 +88,27 @@ describe('permissions service', () => {
...
-    it('calculates 365 days for 1 year licences outside of a leap year', async () => {
+    it('calculates 364 days for 1 year licences outside of a leap year', async () => {
       const startDate = moment('2019-01-01')
       const endDate = await calculateEndDate({ permitId: 'e11b34a0-0c66-e611-80dc-c4346bad0190', startDate: startDate })
-      expect(endDate).toEqual(
-        moment(startDate)
-          .add(365, 'days')
-          .toISOString()
-      )
+      expect(endDate).toEqual(moment(startDate).add(364, 'days').endOf('day').toISOString())
```
jaucourt, line 106: I know you're only modifying what was already here, but given you've got a fixed start date, I'd go for a hard-coded end date rather than calculating it with moment. I think it would be slightly clearer.

packages/sales-api-service/src/services/__tests__/permissions.service.spec.js
```diff
@@ -98,41 +88,27 @@ describe('permissions service', () => {
...
-    it('calculates 366 days for 1 year licences in a leap year', async () => {
+    it('calculates 365 days for 1 year licences in a leap year', async () => {
       const startDate = moment('2020-01-01')
       const endDate = await calculateEndDate({ permitId: 'e11b34a0-0c66-e611-80dc-c4346bad0190', startDate: startDate })
-      expect(endDate).toEqual(
-        moment(startDate)
-          .add(366, 'days')
-          .toISOString()
-      )
+      expect(endDate).toEqual(moment(startDate).add(365, 'days').endOf('day').toISOString())
```
jaucourt, line 111: Same here

packages/sales-api-service/src/services/permissions.service.js
```diff
@@ -31,9 +31,10 @@ export const generatePermissionNumber = async (
   const permit = await getReferenceDataForEntityAndId(Permit, permitId)
 
   const endDate = await calculateEndDateMoment({ permitId, startDate })
-  const endTime = moment(endDate)
-    .add(1, 'hour')
-    .startOf('hour')
+  const endTime =
+    permit.durationMagnitude === 12 && permit.durationDesignator.description === 'M'
+      ? moment(endDate)
+      : moment(endDate).add(1, 'hour').startOf('hour')
```
jaucourt, line 37: I'm not sure we're testing both scenarios here - would you mind checking and adding a test / test case if we're not?

jaucourt (changes requested):

packages/sales-api-service/src/services/permissions.service.js
irisfaraway: @jaucourt From what I can see, both scenarios are covered as we have some tests which use a mock 12-month permit and some which use a mock 1-day permit. I could add some more if you'd prefer them to be more specific though.

packages/sales-api-service/src/services/__tests__/permissions.service.spec.js
irisfaraway: That sounds a lot clearer to me too. Fixed!

packages/sales-api-service/src/services/permissions.service.js
jaucourt: No, that's fine. Thanks for checking :)

jaucourt (approved):

PR #1413, opened 2022-08-02
Welsh not persisting when unauthenticated error

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2992

When on renewals identify page in welsh and trigger unauthenticated licence error page goes back to english following the redirect.


MickStein:
Note while pushing the tests all passed, this may mean tests are missing to check for this issue.

ScottDormand96 (approved):

PR #1421, opened 2022-08-09
Amendment to welsh translation on disability concession page

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3018

Amendment required for welsh translation on both the NI hint and 'no' label.


jaucourt (changes requested):
You've added both the bug and enhancement labels - it needs to be one or the other

MickStein:
@jaucourt removed the bug label

jaucourt (approved):

PR #1422, opened 2022-08-09
Update welsh translation for 'mobile phone'

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3021

The welsh translation for mobile phone is missing on multiple pages. This is set by a label in the cy.json file, so this requires updating to update all pages.


jaucourt (changes requested):
Title again, plus your description should explain why the translation is wrong on multiple pages but you've only changed it on one, e.g. "Translation for 'mobile phone' is wrong on several pages, but it's all set by the same label."

jaucourt (approved):

PR #1430, opened 2022-08-17
Fix easy renewals licence search throwing error when in Welsh

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3013

When searching for a licence and the page is in Welsh the an error is thrown. This needs to be removed so Welsh users can proceed in renewing a licence.


MickStein:
Checked with Naimulla, as this appears to be working. Will close PR.

PR #1431, opened 2022-08-17
Include BST when calculating year-long licence end times

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2889

If a licence ends during BST, we want it to finish at the end of the day in UK time, not in UST.


jaucourt (approved):
Nice work :)

PR #1432, opened 2022-08-19
Fix code smells and broken tests on multibuy branch

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3028

Fixing the two code smells and broken tests on the multibuy branch


irisfaraway (approved):
Looks good to me!

PR #1437, opened 2022-08-25
Add route to view error page

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2987

This feature will allow a user to view the error page by adding '/error' to the url.


MickStein:
Phil provided comments for the test files:

error-routes.spec.js:
- Snapshots should be used to check the http method and path
- Tests for the handler function for rest of tests:
   -- Test output
   -- Passing in mock of h view as jest functions
   -- Assert first argument of handler return is a client error and second argument which will need to mock the values

Routes.spec.js:
- Check routes are there when true and aren’t there when false
- Mock error routes with a symbol
- Mock telesales as a symbol

packages/gafl-webapp-service/src/routes/__tests__/error-routes.spec.js
```diff
@@ -0,0 +1,133 @@
...
+    getLocales: () => []
+  },
+  response: {
+    isBoom: true,
+    output: {
+      statusCode: 400,
+      payload: 'payload'
+    }
+  }
+})
```
jaucourt, line 133: Amend this and routes.spec as we discussed

jaucourt (changes requested):

jaucourt (approved):

jaucourt (changes requested):
Just need to update the README for GAFL

packages/gafl-webapp-service/README.md
```diff
@@ -8,34 +8,37 @@ To run from this directory:
...
+| AIRBRAKE_HOST                     | URL of airbrake host                                             |    no    |                                                           |                               |
+| AIRBRAKE_PROJECT_KEY              | Project key for airbrake logging                                 |    no    |                                                           |                               |
+| SHOW_WELSH_CONTENT                | Display option to change to welsh language                       |    no    |                                                           |                               |
+| ENABLE_ANALYTICS_OPT_IN_DEBUGGING | Set log if analytics been checked in non-production              |    no    |                                                           |                               |
+| ERROR_PAGE_ROUTE                  | Display error pages to support welsh language                    |    no    |                                                           |                               |
```
jaucourt, line 41: Only meant you to add an entry for `ERROR_PAGE_ROUTE` rather than the other two, as they're going to be deleted soon, but we'll leave it as-is for now. Thanks.

jaucourt (approved):

ScottDormand96 (approved):

PR #1461, opened 2022-09-06
Back link on contact page does not work

MickStein:
https://eaflood.atlassian.net/browse/IWTF-2978

This fix will ensure users are able to go back from the contact page by clicking the back button.


MickStein:
I think this has been resolved as this PR is from a while ago. Will close.

PR #1466, opened 2022-09-08
Accessibility - Selected licences title

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3003

Update title on selected licences page

### DIFF
--- a/packages/gafl-webapp-service/src/pages/multibuy/view-licences/view-licences.njk
+++ b/packages/gafl-webapp-service/src/pages/multibuy/view-licences/view-licences.njk
@@ -3,6 +3,8 @@
 {% from "button/macro.njk" import govukButton %}
 {% from "summary-list/macro.njk" import govukSummaryList %}
 
+{% set title = mssgs.multibuy_view_licences_title %}
+{% block pageTitle %}{{ title }}{{ mssgs.gov_uk }}{% endblock %}
 
 {% block content %}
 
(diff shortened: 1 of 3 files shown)


irisfaraway:
@ScottDormand96 Does this also fix the missing title attribute mentioned on Jira or is that a separate PR?

packages/gafl-webapp-service/src/pages/multibuy/view-licences/view-licences.njk
```diff
@@ -3,13 +3,15 @@
...
+{% set title = mssgs.multibuy_view_licences_title %}
+{% block pageTitle %}{{ title }}{{ mssgs.gov_uk }}{% endblock %}
 
 {% block content %}
 
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
-        <h1 class="govuk-heading-l">Selected Licences</h1>
+        <h1 class="govuk-heading-l">Selected licences</h1>
```
jaucourt, line 14: This has been missed when we were moving all hardcoded strings to be labels. Please could you add a new label in en.json and cy.json, then replace the text here with that new label?

jaucourt (changes requested):
Looks fine, but if you could include the new label that'd be great. If it's more work than adding a label, we can create a fresh ticket for the it.

jaucourt (approved):

PR #1476, opened 2022-09-15
Cannot accept/reject analytics on renewal

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3073

Change redirect so for renewal it doesnt redirect to main journey

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
@@ -3,7 +3,7 @@ import analyticsHandler from '../analytics-handler.js'
 import { addLanguageCodeToUri } from '../../processors/uri-helper.js'
 
 jest.mock('../../processors/uri-helper.js', () => ({
-  addLanguageCodeToUri: (_request, uri) => uri
+  addLanguageCodeToUri: jest.fn((_request, uri) => uri)
 }))
 
 jest.mock('../../constants', () => ({
@@ -62,7 +62,7 @@ describe('The analytics handler', () => {
     expect(responseToolkit.redirect).toHaveBeenCalledWith('/buy')
   })
 
-  it.only('calls addLanguageCodeToUri with request and /buy', async () => {
+  it('calls addLanguageCodeToUri with request and /buy', async () => {
     const headers = {
       origin: 'https://localhost:3000',
       referer: 'https://localhost:3000/buy'
--- a/packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
@@ -2,7 +2,9 @@ import { ANALYTICS } from '../../constants.js'
 import analyticsHandler from '../analytics-handler.js'
 import { addLanguageCodeToUri } from '../../processors/uri-helper.js'
 
-jest.mock('../../processors/uri-helper.js')
+jest.mock('../../processors/uri-helper.js', () => ({
+  addLanguageCodeToUri: (_request, uri) => uri
+}))
 
 jest.mock('../../constants', () => ({
   ANALYTICS: {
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -47,5 +47,12 @@ export default async (request, h) => {
       [ANALYTICS.seenMessage]: true
     })
   }
-  return h.redirect(addLanguageCodeToUri(request, '/buy'))
+
+  // https://localhost:3043/buy/name
+  const origin = request.headers.origin
+  const referer = request.headers.referer
+  const redirect = referer.replace(origin, '')
+  console.log('referrer:', redirect)
```
jaucourt, line 55: remove the console log

packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -47,5 +47,12 @@ export default async (request, h) => {
       [ANALYTICS.seenMessage]: true
     })
   }
-  return h.redirect(addLanguageCodeToUri(request, '/buy'))
+
+  // https://localhost:3043/buy/name
```
jaucourt, line 51: this can be deleted

packages/gafl-webapp-service/src/handlers/analytics-handler.js
irisfaraway, line 55: Do we want to keep this logging or was it just for debug purposes?

packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -47,5 +47,12 @@ export default async (request, h) => {
...
+  const origin = request.headers.origin
+  const referer = request.headers.referer
+  const redirect = referer.replace(origin, '')
```
jaucourt, line 54: We need to check that we're redirecting to a valid location - so if the host / port match the host / port of the current page, then redirect to the path after host / port

For example:
Referer is https://localhost:3043/buy/name and the current url is https://localhost:3043/buy/process-analytics-preferences, we can redirect to `/buy/name`

However, if the referer is `http://some.shady.url:9999/who/knows/what`, we should redirect to `/buy`

jaucourt (changes requested):

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -17,35 +17,41 @@ describe('The analytics handler', () => {
...
+  it.each([
+    ['https://localhost:3000', 'https://localhost:3000/buy/name', '/buy/name'],
+    ['https://localhost:1234', 'https://localhost:1234/example/test/redirect', '/example/test/redirect'],
+    ['https://testserver-example-fish', 'https://testserver-example-fish/buy/renew/identify', '/buy/renew/identify']
+  ])('redirects to correct page', async (origin, referer, redirect) => {
     const payload = {
       analyticsResponse: 'accept'
     }
-    const request = generateRequestMock(payload)
+    const request = generateRequestMock(payload, 'analytics', origin, referer)
```
irisfaraway, line 28: I notice in these tests we sometimes have `payload` and `analytics` in single quotes and sometimes not – is there a reason we alternate?

irisfaraway (commented):
Some possibly silly questions!

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -18,19 +18,38 @@ describe('The analytics handler', () => {
...
+    ['https://testserver-example-fish', 'https://testserver-example-fish/buy/renew/identify', 'https://hdfhdskfhs-ghj-vgjh']
+  ])('redirects to correct page if url matches header', async (origin, referer, host) => {
+    const payload = {
+      analyticsResponse: 'accept'
+    }
+    const request = generateRequestMock(payload, 'analytics', origin, referer, host)
+    const responseToolkit = generateResponseToolkitMock()
+    await analyticsHandler(request, responseToolkit)
+    expect(responseToolkit.redirect).toHaveBeenCalledWith(addLanguageCodeToUri(request, '/buy'))
+  })
```
jaucourt, line 51: This test and the previous one both have the same name, 'redirects to correct page if url matches header'. I think you need to describe them both better. In this second set of tests it looks like you're testing the default, so state that in the test name - something like 'redirects to /buy if the HTTP_REFERER host and port does not match the host and port of the current page'

Furthermore, I think the `addLanguageCodeToUri` is muddying the water. You should mock `addLanguageCodeToUri` so it returns whatever is sent to it without modifying it, then your assertion could omit `addLanguageCodeToUri` entirely and do `toHaveBeenCalledWith('/buy')`. Do this for the first test case too. We're not testing addLanguageCodeToUri, so it only causes confusion.

jaucourt (changes requested):

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -26,28 +26,30 @@ describe('The analytics handler', () => {
       'https://testserver-example-fish',
       '/buy/renew/identify'
     ]
-  ])('redirects to correct page if url matches header', async (origin, referer, host, redirect) => {
+  ])('redirects to correct page if the HTTP_REFERER host matchesthe host of the current page', async (origin, referer, host, redirect) => {
```
jaucourt, line 29: typo - missing space between 'matches' and 'host'

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -26,28 +26,30 @@ describe('The analytics handler', () => {
...
     const payload = {
       analyticsResponse: 'accept'
     }
     const request = generateRequestMock(payload, 'analytics', origin, referer, host)
     const responseToolkit = generateResponseToolkitMock()
+    addLanguageCodeToUri.mockReturnValueOnce(redirect)
```
jaucourt, line 35: This will make it so your test always passes - no matter what's passed to addLanguageCodeToUri. Outside of this function, at the top of the file, do:
```
jest.mock('../../processors/uri-helper.js', () => ({
  addLanguageCodeToUri: uri => uri
}))
```

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -26,28 +26,30 @@ describe('The analytics handler', () => {
...
     ['https://localhost:1234', 'https://localhost:1234/example/test/redirect', 'https://notsamehost:1234'],
     ['https://testserver-example-fish', 'https://testserver-example-fish/buy/renew/identify', 'https://hdfhdskfhs-ghj-vgjh']
-  ])('redirects to correct page if url matches header', async (origin, referer, host) => {
+  ])('redirects to /buy if the HTTP_REFERER host does not match the host of the current page', async (origin, referer, host) => {
     const payload = {
       analyticsResponse: 'accept'
     }
     const request = generateRequestMock(payload, 'analytics', origin, referer, host)
     const responseToolkit = generateResponseToolkitMock()
+    addLanguageCodeToUri.mockReturnValueOnce('/buy')
```
jaucourt, line 50: Remove this once you've implemented the mock for addLanguageCodeToUri I've described above

jaucourt (changes requested):

jaucourt (approved):

PR #1477, opened 2022-09-15
analytics banner only /buy

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3075

Only display analytics banner when uri begins with /buy

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/page-handler.js
+++ b/packages/gafl-webapp-service/src/handlers/page-handler.js
@@ -116,10 +116,7 @@ export default (path, view, completion, getData) => ({
     pageData.analyticsSelected = analytics[ANALYTICS.selected]
     pageData.acceptedTracking = analytics[ANALYTICS.acceptTracking]
 
-    pageData.displayAnalytics = false
-    if (request.path.startsWith('/buy')) {
-      pageData.displayAnalytics = true
-    }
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -116,6 +116,11 @@ export default (path, view, completion, getData) => ({
     pageData.analyticsSelected = analytics[ANALYTICS.selected]
     pageData.acceptedTracking = analytics[ANALYTICS.acceptTracking]
 
+    pageData.displayAnalytics = false
+    if (request.path.startsWith('/buy')) {
```
irisfaraway, line 120: Could we add a test for this?

irisfaraway (dismissed):

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -116,6 +116,11 @@ export default (path, view, completion, getData) => ({
...
+      pageData.displayAnalytics = true
+    }
```
jaucourt, line 122: This would be more concise if you did 
```
pageData.displayAnalytics = request.path.startsWith('/buy')
```

jaucourt (changes requested):

jaucourt (approved):

PR #1482, opened 2022-09-20
Add route to view error page for a server error

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3049

Should a server side error occur (500 error) the user will see a page describing the error.


MickStein:
Deleted to restart as the client error ticket has been added to develop and the merge has caused issues.

PR #1485, opened 2022-09-22
Use chosen language when moving from renewals to new licence journey

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2899

The "get a new fishing licence" link from the renewals journey was not persisting the user's chosen language - for example, if the user had previously selected Welsh, the link swapped them back to English. This PR fixes that.

It also includes some test refactoring to break up some of the tests to be a bit more granular.


irisfaraway:
I am not happy with the mocking situation here but it's what I've ended up with. Feedback from reviewers would be very welcome!

jaucourt (approved):

PR #1494, opened 2022-10-04
Identical Licences

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3057

Adding a warning to the service incase indentical licences based on name, type, length and start date


packages/gafl-webapp-service/src/locales/cy.json
```diff
@@ -393,6 +393,7 @@
   "licence_type_title_you": "Pa fath o drwydded hoffech chi ei chael?",
   "licence_type_troute_three_rod": "Dim ond trwydded tair gwialen y bydd ei hangen arnoch os ydych yn bwriadu pysgota â thair gwialen ar yr un pryd. Gall hyn fod ar gyfer pysgota cerpynnod arbenigol, er enghraifft. Bydd hyn yn addas ar gyfer pob math o bysgota â gwialen ac yn gadael i chi bysgota brithyllod anfudol a physgod bras, yn amodol ar y ",
   "licence_type_troute_two_rod": "Y drwydded pysgod bras â dwy wialen yw’r drwydded wialen symlaf. Bydd hyn yn eich gwarchod ar gyfer pob math o bysgota gyda gwialen ac yn gadael i chi bysgota brithyllod anfudol a physgod bras. Bydd yn caniatáu i chi bysgota gydag un neu ddwy wialen ar yr un pryd, yn amodol ar y ",
+  "multibuy_licences_warning_duplicate": "You currently have multiple versions of the same licence in your basket. You can change or remove licences, or select Continue to proceed without change",
```
MickStein, line 396: does the 'Continue' need to be in apostrophes?

packages/gafl-webapp-service/src/pages/multibuy/view-licences/route.js
```diff
@@ -18,7 +19,9 @@ export const getData = async request => {
     index
   }))
 
-  return { licences }
+  const duplicate = await checkDuplicates(licences)
+
+  return { licences, duplicate }
```
MickStein, line 24: Does a test need to be done with this change to check the return value?

packages/gafl-webapp-service/src/handlers/multibuy-duplicate-handler.js
```diff
@@ -0,0 +1,23 @@
...
+
+  let checkDuplicates = false
+  for (let licence = 0; licence < licencesCheck.length; licence++) {
+    for (let licenceCompare = licence + 1; licenceCompare < licencesCheck.length; licenceCompare++) {
+      if (_.isEqual(licencesCheck[licence], licencesCheck[licenceCompare])) {
+        checkDuplicates = true
+        break
+      }
+    }
+  }
```
jaucourt, line 20: I think this is a bit complex. Essentially, you need a hash that allows for quick comparison between elements. If it's an identical licenceHolder, type, length and start date that constitute a duplicate, you can create a 'comparison array', by doing:
```
const hasDuplicates = licencesArray => {
  const haystack = licencesArray.map(licence => `${licence.licenceHolder}${licence.type}${licence.length}${licence.start}`);
  const duplicates = haystack.filter(needle => { 
      return haystack.filter(potentialNeedle => needle === potentialNeedle).length > 1
  });
  return duplicates.length > 1
}
```
There are a few other good examples here: https://stackoverflow.com/questions/30735465/how-can-i-check-if-the-array-of-objects-have-duplicate-property-values, I think using a Set might be a good option although I'm not sure that it's supported in the version of Node we're on:
```
let seen = new Set();
var hasDuplicates = values.some(function(currentObject) {
    return seen.size === seen.add(currentObject.name).size;
});
```

jaucourt (changes requested):
Sorry for the number of comments, I can see you've had a good stab at it

packages/gafl-webapp-service/src/handlers/__tests__/multibuy-duplicate-handler.spec.js
```diff
@@ -1,152 +1,40 @@
...
-        length: '12M',
-        start: '2025-01-01',
-        price: '30'
-      }
-    ]
-
-    const result = await checkDuplicates(licences)
+  it('identifies when array has no duplicate licences', async () => {
+    const licenses = [getLicenceOne(), getLicenceTwo(), getLicenceThree()]
+    const result = await hasDuplicates(licenses)
```
jaucourt, line 37: I think you've misunderstood what I meant when I said 'there should be two tests'. I mentioned using it.each, because you were writing the same test several times. So two tests, one to check arrays with duplicates and one to check arrays without duplicates. Then several different _cases_ passed to each test - I gave three licence holders as an example, but I'm not sure that's enough. What do you think? Do you feel it's well tested as is?

ScottDormand96: I think when duplicate one test would be enough, but I will add multiple tests for when not duplicate based on if one area not matching (licence holder type etc)

jaucourt (approved):

packages/gafl-webapp-service/src/handlers/multibuy-duplicate-handler.js
```diff
@@ -0,0 +1,7 @@
+export const hasDuplicates = async licencesArray => {
+  const licences = licencesArray.map(licence => `${licence.licenceHolder}${licence.type}${licence.length}${licence.start}`)
+  const duplicates = licences.filter(licenceChecking => {
+    return licences.filter(licenceComparing => licenceChecking === licenceComparing).length > 1
+  })
+  return duplicates.length > 1
+}
```
jaucourt, line 7: Just one last thing. I think this is more a processor than handler, so please could you switch this into processors (renaming it to `multibuy-processor.js` in the process), thanks

jaucourt (changes requested):
Just one last thing </colomboVoice>

PR #1495, opened 2022-10-04
Google search takes user incorrect page

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3070

Adding a redirect to the BOBO page if not already been through journey


jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/concessions/disability/disability-concession.njk
```diff
@@ -1,4 +1,5 @@
 {% extends "standard-form.njk" %}
+
```
jaucourt, line 2: Doing this in quite a few pages, reset these to the state they were in to reduce the number of files in this PR

packages/gafl-webapp-service/src/pages/contact/address/select/address-select.njk
```diff
@@ -1,5 +1,7 @@
 {% extends "standard-form.njk" %}
 
+<meta name="robots" content="noindex">
```
jaucourt, line 3: This shouldn't be here, I think

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/route.js
```diff
@@ -14,8 +15,20 @@ export const validator = payload => {
...
+    console.log('ghgj')
+    throw new GetDataRedirect(LICENCE_FOR.uri)
+  }
+}
+
 export const getData = async request => {
   const { isLicenceForYou } = await request.cache().helpers.transaction.getCurrentPermission()
+  const status = await request.cache().helpers.status.getCurrentPermission()
+
+  console.log(status)
```
jaucourt, line 30: Console log here and directly above that shouldn't be here

packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
```diff
@@ -235,6 +237,35 @@ describe('The page handler function', () => {
...
+    const toolkit = getMockToolkit()
+    const mockRequest = getMockRequest(null, pageUri)
+    await get(mockRequest, toolkit)
+    const pageData = toolkit.view.mock.calls[0][1]
+    expect(pageData).toEqual(
+      expect.objectContaining({
+        notJourneyBeginning: false
+      })
+    )
+  })
```
jaucourt, line 268: I think you've misunderstood me. We shoudn't have a value that we have to set on every page. If you reverse this to `journeyBeginning`, which gets set to true on the licence-for and renewal id pages, then in the layout.njk it can show the robots meta tag when that's not true. What you're doing, whilst it will work, means setting the value on every single page other than the two I've just mentioned - obviously far more onerous.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/__tests__/route.spec.js
```diff
@@ -1,31 +1,73 @@
...
+  describe('redirectToStartOfJourney', () => {
+    it('should throw a redirect if not been to LICENCE_FOR page', async () => {
+      const transaction = () => ({
+        isLicenceForYou: true
+      })
+      const status = () => ({
+        [LICENCE_FOR.page]: false
+      })
+      const func = async () => await getData(mockRequest(status, transaction))
+      await expect(func).rejects.toThrow(new GetDataRedirect(LICENCE_FOR.uri))
```
jaucourt, line 59: Until I get around to adding that new matcher, this would work just as well if you did:
```
await expect(func).rejects.toThrow(GetDataRedirect)
```
I'll leave it up to you whether you change it, just don't assume that you're testing that the thrown error has `redirectUrl` set to `LICENCE_FOR.uri`. You could change `redirectToStartOfJourney` to be:
```
const redirectToStartOfJourney = status => {
  if (!status[LICENCE_FOR.page]) {
     throw new GetDataRedirect('any old rubbish')
  }
 }
```
and your test would still pass.

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/__tests__/route.spec.js
```diff
@@ -1,31 +1,73 @@
...
+
+  describe('redirectToStartOfJourney', () => {
+    it('should throw a redirect if not been to LICENCE_FOR page', async () => {
+      const transaction = () => ({
+        isLicenceForYou: true
+      })
+      const status = () => ({
+        [LICENCE_FOR.page]: false
+      })
+      const func = async () => await getData(mockRequest(status, transaction))
```
jaucourt, line 58: Another comment - you don't need the async and await here. It'd work exactly the same if you did:
```
const func = () => getData(mockRequest(status, transaction))
```
The reason for this is that `func` would still return a Promise, which would then be `await`ed by the expect on the following line. So long as there's an `await` in there somewhere, it'll work fine

jaucourt (changes requested):
A couple of comments, I'll leave it up to you whether you want to implement the suggested changes in them as your test works without it. The only one that needs to be implemented is where you test for the error not being thrown, as that doesn't actually test what you're expecting it to.

jaucourt (approved):

MickStein (approved):

PR #1496, opened 2022-10-05
View licence page missing '30 minutes after payment' in licence start

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3002

The value for 'Start date' on the view licence page should display '30 minutes after payment' when this option has been selected.


MickStein:
This needs to come from the multibuy branch, which I have not started from. Due to this I will close this one and make a new branch.

PR #1498, opened 2022-10-06
Add route to view error page for a server error

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3049

Should a server side error occur (500 error) the user will see a page describing the error.


MickStein:
I think this has been completed with Scott so will close PR.

PR #1503, opened 2022-10-13
Welsh language - persist on error page

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-2917

Welsh language persist when an erro page loads up


MickStein (approved):

packages/gafl-webapp-service/src/session-cache/__tests__/session-manager.spec.js
```diff
@@ -116,3 +122,67 @@ describe('The user', () => {
...
+    state: () => {},
+    redirect: jest.fn(() => ({
+      takeover
+    }))
+  })
+  const initialisedSessionManager = sessionManager(sessionCookieName)
+
+  beforeEach(jest.clearAllMocks)
+
+  it("redirects to agreed handler when status is already agreed and request path isn't in exempt set", async () => {
```
irisfaraway, line 155:
```suggestion
  it('redirects to agreed handler when status is already agreed and request path isn't in exempt set', async () => {
```

I know I'm nitpicking here but :)

packages/gafl-webapp-service/src/session-cache/__tests__/session-manager.spec.js
```diff
@@ -116,3 +122,67 @@ describe('The user', () => {
...
   })
 })
+
+describe('takeover by agreed handler once agreed flag is set', () => {
+  const sessionCookieName = 'garibaldi'
+  const getMockRequest = () => ({
+    path: 'crazy-paving',
+    state: {
+      [sessionCookieName]: {
+        id: 'a disappointing currant embedded biscuit'
```
irisfaraway, line 132: 😂

irisfaraway (changes requested):
One very minor style thing, other than that LGTM

packages/gafl-webapp-service/src/session-cache/__tests__/session-manager.spec.js
jaucourt: I would normally use single quotes, but note the apostrophe in isn't

irisfaraway: Oops. You're totally right!

irisfaraway (approved):

PR #1510, opened 2022-10-20
Fix for back link button

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3109

User should be able to click the back button and go back to relevant page, at present it throws a 500 error.

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/page-handler.js
+++ b/packages/gafl-webapp-service/src/handlers/page-handler.js
@@ -43,9 +43,9 @@ const getBackReference = async (request, view) => {
   if (!current || !current.backLink) {
     return null
   }
-
+  console.log(current)
   if (typeof current.backLink === 'function') {
-    const backLink = current.backLink(
+    const backLink = await current.backLink(
       await request.cache().helpers.status.getCurrentPermission(),
       await request.cache().helpers.transaction.getCurrentPermission()
     )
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -43,9 +43,9 @@ const getBackReference = async (request, view) => {
   if (!current || !current.backLink) {
     return null
   }
-
+  console.log(current)
```
jaucourt, line 46: I don't think this is meant to be here

jaucourt (changes requested):

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -43,9 +43,8 @@ const getBackReference = async (request, view) => {
   if (!current || !current.backLink) {
     return null
   }
-
   if (typeof current.backLink === 'function') {
-    const backLink = current.backLink(
+    const backLink = await current.backLink(
```
jaucourt, line 47: When you come across a bug, best practice is that you should write a test that demonstrates the bug (i.e. fails), then write code to make that test pass. That way, the bug will never occur again, as there's a test that catches it.

jaucourt (changes requested):

jaucourt (approved):

PR #1514, opened 2022-10-24
Demo of redirectWithLanguageCode function

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3027

This PR exists to demonstrate the usage of a new `redirectWithLanguageCode` function on the response toolkit. This function allows us to call a redirect for the correct language without having to also call `addLanguageCodeToUri` each time.

As an example of how this would impact the code and tests, I added it to the agreed-handler, since the tests for that are pretty gnarly.


irisfaraway:
This has been superseded by #1580

PR #1524, opened 2022-10-31
Start after 30 minutes text not showing with additional licences

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3002

When the user adds additional licences with future start dates, the start time text for "30 minutes after" should remain for previous licences.

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
@@ -1,9 +1,9 @@
 import { displayStartTime, displayEndTime, displayExpiryDate, advancePurchaseDateMoment } from '../date-and-time-display.js'
 import moment from 'moment-timezone'
-import constants from '@defra-fish/business-rules-lib'
+import constant from '../../../../business-rules-lib/src/constants.js'
 
-jest.mock('@defra-fish/business-rules-lib', () => ({
-  START_AFTER_PAYMENT_MINUTES: 30,
+jest.mock('../../../../business-rules-lib/src/constants.js', () => ({
+  START_AFTER_PAYMENT_MINUTES: 40,
   SERVICE_LOCAL_TIME: 'Europe/London'
 }))
 
@@ -13,7 +13,9 @@ jest.mock('moment-timezone', () => ({
     add: () => ({
       format: () => '2020-01-06T00:00:00.000',
       locale: () => ({ format: () => ({ replace: () => {} }) })
-    })
+    }),
+    locale: jest.fn(() => ({ format: () => '' })),
+    format: () => {}
   }),
   format: () => {},
   locale: jest.fn(),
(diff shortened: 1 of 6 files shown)


packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +43,39 @@ describe('displayStartTime', () => {
     expect(moment.utc).toHaveBeenCalledWith(startDate, null, expect.any(String))
   })
 
+  it.each([[50], [10000], [0]])(
+    'expected text replaces the date string if start after payment is selected using correct constant',
```
jaucourt, line 47: It's good to include a parameter within the test description so that you can see the difference between the tests in the output. You can include it by putting %s into the test description, for example 'generates a string of payment delay (%s) followed by licence_summary_minutes_after_payment label when licence is to start after payment'

Also, you're varying the `START_AFTER_PAYMENT_MINUTES` constant, you should put `startDate` and the `licence_summary_minutes_after_payment` message in there too and vary them a bit

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +43,39 @@ describe('displayStartTime', () => {
...
+      const mssgs = getSampleRequest().i18n.getCatalog()
+      const expectedReturnText = constant.START_AFTER_PAYMENT_MINUTES + mssgs.licence_summary_minutes_after_payment
+      moment.utc.mockImplementation(() => ({
+        tz: () => ({
+          locale: jest.fn(() => ({ format: () => '' })),
+          format: () => {}
+        })
+      }))
+      const startAfterPaymentMinutes = displayStartTime(getSampleRequest(), { startDate, licenceToStart: 'after-payment' })
+      expect(startAfterPaymentMinutes).toEqual(expectedReturnText)
```
jaucourt, line 60: This, and the following test, are both incredibly complicated, especially considering what you're testing. 
1) Why are you adding a new mock for moment, that is identical to the mock you've created on lines 16-18?
2) Why are you getting a sample request twice? Create a `mockRequest` at the start of the test

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +43,39 @@ describe('displayStartTime', () => {
     expect(moment.utc).toHaveBeenCalledWith(startDate, null, expect.any(String))
   })
 
+  it.each([[50], [10000], [0]])(
+    'expected text replaces the date string if start after payment is selected using correct constant',
+    async startAfterMinutes => {
+      constant.START_AFTER_PAYMENT_MINUTES = startAfterMinutes
+      const startDate = '2021-01-01T00:00:00.000Z'
+      const mssgs = getSampleRequest().i18n.getCatalog()
+      const expectedReturnText = constant.START_AFTER_PAYMENT_MINUTES + mssgs.licence_summary_minutes_after_payment
```
jaucourt, line 52: It's preferable to use template quotes rather than the concatenation operator, e.g.
```
const expectedReturnText = `${constant.START_AFTER_PAYMENT_MINUTES}${mssgs.licence_summary_minutes_after_payment}
```
Also, for clarity, use the passed-in variable rather than what you've assigned it to:
```
const expectedReturnText = `${startAfterMinutes}${mssgs.licence_summary_minutes_after_payment}
```
Better still, hard-code the expected return text in a case variable, like in "displays the date in the correct format, when permission date is %s and current date and time is %s" below. Avoid too much logic in tests, otherwise you'll get to a point you need a test for your test!

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +43,39 @@ describe('displayStartTime', () => {
...
+    moment.utc.mockReturnValueOnce(realMoment(startDate))
+    moment.utc.mockImplementation(() => ({
+      tz: () => ({
+        locale: jest.fn(() => ({ format: () => '' })),
+        format: () => {}
+      })
+    }))
+    const returnDateValue = displayStartTime(getSampleRequest(), { startDate, licenceToStart: 'not-after-payment' })
+    expect(returnDateValue).toEqual(expectedReturnDateValue)
+  })
```
jaucourt, line 77: I don't think this is necessary - what does it test that the following test doesn't?

packages/gafl-webapp-service/src/processors/date-and-time-display.js
```diff
@@ -23,6 +26,11 @@ export const displayStartTime = (request, permission) => {
     .format('h:mma')
     .replace('12:00am', mssgs.licence_start_time_am_text_0)
     .replace('12:00pm', mssgs.licence_start_time_am_text_12)
+
+  if (permission.licenceToStart === licenceToStart.AFTER_PAYMENT) {
+    return START_AFTER_PAYMENT_MINUTES + startAfter30Text
```
jaucourt, line 31: Better to use template quotes rather than the concatenation operator

jaucourt (changes requested):

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +48,26 @@ describe('displayStartTime', () => {
...
+  ])(
+    'generates a string of payment delay (%s) followed by licence_summary_minutes_after_payment label when licence is to start after payment',
+    async (startAfterMinutes, date, message) => {
+      constant.START_AFTER_PAYMENT_MINUTES = startAfterMinutes
+      const mockRequest = getSampleRequest({
+        licence_summary_minutes_after_payment: message
+      })
+      const mssgs = mockRequest.i18n.getCatalog()
+      console.log(mssgs)
+      const startDate = date
```
jaucourt, line 64: Not needed - just change 'date' to 'startDate' on line 57

jaucourt (changes requested):

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -1,25 +1,39 @@
...
+  utc: jest.fn(() => ({
+    tz: () => ({
+      locale: jest.fn(() => ({ format: () => '' })),
+      format: () => {}
+    })
+  }))
 }))
 
-const getSampleRequest = () => ({
+const getSampleRequest = (overrides = {}) => ({
```
jaucourt, line 30: change this to `labelOverrides` so that it's clear

jaucourt, line 30: change this to labelOverrides so that it's clear

jaucourt (changes requested):

packages/gafl-webapp-service/src/processors/__tests__/date-and-time-display.spec.js
```diff
@@ -34,6 +48,23 @@ describe('displayStartTime', () => {
...
+    [[10000], ['2010-12-23T14:00:00.000Z'], [' minutes until the salmon are yours']],
+    [[0], ['2029-06-030T01:00:00.000Z'], [' minutes until cobwebs are brushed off your rod']]
+  ])(
+    'generates a string of payment delay (%s) followed by licence_summary_minutes_after_payment label when licence is to start after payment',
+    async (startAfterMinutes, startDate, message) => {
+      constant.START_AFTER_PAYMENT_MINUTES = startAfterMinutes
+      const mockRequest = getSampleRequest({
+        licence_summary_minutes_after_payment: message
+      })
+      const expectedReturnText = `${constant.START_AFTER_PAYMENT_MINUTES}${message}`
```
jaucourt, line 62: You've done half of what I said. Please do the other half

jaucourt (changes requested):

jaucourt (approved):

PR #1526, opened 2022-11-02
Contact Page shows incorrect information for license confirmation page

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3103

When a user purchases a 12 month licence and selects email for notifications and text for reminders both should be reflect as the chosen contact options.


packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
```diff
@@ -101,7 +112,11 @@ class RowGenerator {
   }
 
   generateContactRow (label, href, visuallyHiddenText, id, contactTextSpec = CONTACT_TEXT_DEFAULT) {
-    return this._generateRow(this.labels[label], this._getContactText(contactTextSpec), href, this.labels[visuallyHiddenText], id)
+    const contactText =
+      label !== 'contact_summary_row_contact'
```
jaucourt, line 116: Reverse the logic here - it's generally best practice to go for `if (x === y) { /* stuff */ } else { /*other stuff */ }`

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -141,6 +142,46 @@ describe('contact-summary > route', () => {
...
+          licensee: {
+            ...addressAndContact,
+            postalFulfilment: true,
+            preferredMethodOfConfirmation: 'Email',
+            preferredMethodOfReminder: 'Text',
+            mobilePhone: '07700900900',
+            email: 'new3@example.com',
+            preferredMethodOfNewsletter: 'Yes'
+          }
+        }
```
jaucourt, line 177: change to using a mock generator method

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -64,318 +101,190 @@ const generateRequestMock = (permission, query, status = {}) => ({
...
-      }
-      const sampleRequest = generateRequestMock(permission)
-
-      const { summaryTable } = await getData(sampleRequest)
-
-      expect(summaryTable).toMatchSnapshot()
+  it.each([
+    [HOW_CONTACTED.none, mssgs.no],
+    [null, mssgs.yes]
+  ])('returns yes or no to newsletter depending on user preference', async (preferredMethodOfNewsletter, expectedReturn) => {
```
jaucourt, line 232: A small point, but rather than covering both cases in the test name, it should be tailored to provide a unique test name depending on the parameters that're being passed to the test. So we'd have 'newsletter is not sent when preferredMethodOfNewsletter is HOW_CONTACTED.none' and 'newsletter is sent when preferredMethodOfNewsletter is null'. 

You could do that by amending to this:
```
  it.each([
    { preferredMethodOfNewsletter: HOW_CONTACTED.none, expectedReturn: mssgs.no, action: 'not sent' },
    { preferredMethodOfNewsletter: null, expectedReturn: mssgs.yes, action: 'sent' }
  ])('newsletter is $action when preferredMethodOfNewletter is $preferredMethodOfNewsletter', async ({ preferredMethodOfNewsletter, expectedReturn }) => {
    const samplePermission = getMockPermission({
      preferredMethodOfNewsletter
    })
    const { summaryTable } = await getData(
      getRequestMock({
        permission: samplePermission
      })
    )
    expect(summaryTable[4].value.text).toBe(expectedReturn)
  })

```
Also, given that the second test description ends up as 'newsletter is sent when preferredMethodOfNewsletter is null', is that right? It doesn't sound right. I would have thought it should be sent when `preferredMethodOfNewsletter` is `HOW_CONTACTED.email`

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -64,318 +101,190 @@ const generateRequestMock = (permission, query, status = {}) => ({
...
       const status = {
         [LICENCE_CONFIRMATION_METHOD.page]: false
       }
-      const permission = { licenceLength: '12M', isRenewal: true, licensee: {} }
-      const mockRequest = generateRequestMock(permission, '', status)
-
+      const mockRequest = getRequestMock({ status })
       await expect(() => getData(mockRequest)).rejects.toThrowRedirectTo(LICENCE_CONFIRMATION_METHOD.uri)
     })
   })
```
jaucourt, line 290: I've noticed that in contact-summary/route.js, there is the following:
```
  status.fromSummary = CONTACT_SUMMARY_SEEN
  await request.cache().helpers.status.setCurrentPermission(status)
```
As far as I can see, this isn't ever tested. Please can you add a couple of tests that fromSummary gets set and that request.cache().helpers.status.setCurrentPermission is called with the status object?

jaucourt (changes requested):
Good work here, just a couple of final things

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -33,349 +36,274 @@ const route = require('../route.js').default
...
+          fromSummary: false,
           [ADDRESS_ENTRY.page]: true,
           [ADDRESS_SELECT.page]: true,
           [CONTACT.page]: true,
           [LICENCE_FULFILMENT.page]: true,
           [LICENCE_CONFIRMATION_METHOD.page]: true,
           ...status
         }),
-        setCurrentPermission: () => {}
+        setCurrentPermission
```
jaucourt, line 98: I think the parameter should be called `setStatusPermission`, even though this will mean a less succinct reference here: `setCurrentPermission: setStatusPermission`.

Also, in every test where you're not at all interested in `status.setCurrentPermission`, because you've not set a default, you have to specify a value for it. Set the default as `() => {}`, then you can remove the specification of `setCurrentPermission` in all tests not interested in it.

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
jaucourt: In general execution, `preferredMethodOfNewsletter` would never be null. So in place of `{ preferredMethodOfNewsletter: null, expectedReturn: mssgs.yes, action: 'sent' }`, do the following cases:
```
{ preferredMethodOfNewsletter: HOW_CONTACTED.email, expectedReturn: mssgs.yes, action: 'sent' }
{ preferredMethodOfNewsletter: HOW_CONTACTED.text, expectedReturn: mssgs.yes, action: 'sent' }
{ preferredMethodOfNewsletter: HOW_CONTACTED.post, expectedReturn: mssgs.yes, action: 'sent' }
```
Tests form 'use cases' for the code, and although it's possible to pass 'null' for `preferredMethodOfNewsletter`, it wouldn't ever be set to null. It's more helpful to set actual, expected values that show the use of the code in expected operation.

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -33,349 +36,274 @@ const route = require('../route.js').default
...
-          }
-        }
-        const sampleRequest = generateRequestMock(permission)
-
-        const { summaryTable } = await getData(sampleRequest)
-
-        expect(summaryTable).toMatchSnapshot()
+  it.each([
+    { preferredMethodOfNewsletter: HOW_CONTACTED.none, expectedReturn: mssgs.no, action: 'not sent' },
+    { preferredMethodOfNewsletter: null, expectedReturn: mssgs.yes, action: 'sent' }
```
jaucourt, line 157: This test is really unclear. The title of the test is `'newsletter is $action when preferredMethodOfNewletter is $preferredMethodOfNewsletter'`, but in the test you don't assert on, and in fact don't even use `action`. Did you mean to?

If you are meaning to test on expectedReturn, what you've done isn't great practice. What you've done immediately above in `'changeLicenceDetails should be %s when isLicenceForYou is %s'` is much better, where you set the message in the catalog within the test, then assert that this is being used in the correct way in the function result.

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -33,349 +36,274 @@ const route = require('../route.js').default
...
-    })
+    getCatalog: () => catalog
   },
   query,
   url: {
     search: ''
   }
 })
 
+const mssgs = getRequestMock().i18n.getCatalog()
```
jaucourt, line 114: This is only used in one test, so shouldn't be here.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
MickStein: that makes alot more sense thanks!

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -33,349 +36,278 @@ const route = require('../route.js').default
...
-        }
-        const sampleRequest = generateRequestMock(permission)
-
-        const { summaryTable } = await getData(sampleRequest)
-
-        expect(summaryTable).toMatchSnapshot()
+  it.each([
+    { preferredMethodOfNewsletter: HOW_CONTACTED.email, expectedReturn: mssgs.yes, action: 'sent' },
+    { preferredMethodOfNewsletter: HOW_CONTACTED.text, expectedReturn: mssgs.yes, action: 'sent' },
+    { preferredMethodOfNewsletter: HOW_CONTACTED.post, expectedReturn: mssgs.yes, action: 'sent' }
```
jaucourt, line 163: It looks like you've removed the only case that tested the reverse (`[HOW_CONTACTED.none, mssgs.no]`), so now `expectedReturn` and `action` don't vary at all. You needed to keep that original case, so put it back in. Then check that you're not passing in something that never varies - if you are that can be hard-coded into the test.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
MickStein: Unsure why these have no updated sorry, as these are out of date.

packages/gafl-webapp-service/src/pages/summary/contact-summary/__tests__/route.spec.js
```diff
@@ -158,23 +156,23 @@ describe('contact-summary > route', () => {
...
-    { preferredMethodOfNewsletter: HOW_CONTACTED.email, expectedReturn: mssgs.yes, action: 'sent' },
-    { preferredMethodOfNewsletter: HOW_CONTACTED.text, expectedReturn: mssgs.yes, action: 'sent' },
-    { preferredMethodOfNewsletter: HOW_CONTACTED.post, expectedReturn: mssgs.yes, action: 'sent' }
-  ])(
-    'newsletter is $action when preferredMethodOfNewletter is $preferredMethodOfNewsletter',
-    async ({ preferredMethodOfNewsletter, expectedReturn }) => {
-      const samplePermission = getMockPermission({
-        preferredMethodOfNewsletter
+    ['aye', HOW_CONTACTED.email, 'yes'],
+    ['negative, Ghost Rider', HOW_CONTACTED.none, 'no']
```
jaucourt, line 160: Please will you change the 'aye' and 'negative, Ghost Rider' to be something that's not in the mock catalog already. Then we're done!

jaucourt (changes requested):

jaucourt (approved):
Thanks for all the work that's gone into this

PR #1534, opened 2022-11-09
Welsh - contact details

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3119

Minor amendments for check contact details none bobo


packages/gafl-webapp-service/src/pages/summary/contact-summary/contact-summary.njk
```diff
@@ -44,10 +44,16 @@
...
             {{ csrf() }}
         </form>
-
-        <p class="govuk-body-m">
-            <a class="govuk-link" href="{{ data.uri.licenceSummary }}">{{ mssgs.change_licence_details }}</a>
-        </p>
+        
+        {% if data.isLicenceForYou %}
+            <p class="govuk-body-m">
+                <a class="govuk-link" href="{{ data.uri.licenceSummary }}">{{ mssgs.change_licence_details_you }}</a>
```
MickStein, line 50: Are you able to make this a one liner like:
`{% set title = mssgs.dob_title_you if data.isLicenceForYou else mssgs.dob_title_other %}
` ?

MickStein (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/contact-summary.njk
```diff
@@ -44,9 +44,9 @@
 
             {{ csrf() }}
         </form>
-
+        
         <p class="govuk-body-m">
-            <a class="govuk-link" href="{{ data.uri.licenceSummary }}">{{ mssgs.change_licence_details }}</a>
+            <a class="govuk-link" href="{{ data.uri.licenceSummary }}">{{ mssgs.change_licence_details_you if data.isLicenceForYou else change_licence_details_other }}</a>
```
jaucourt, line 49: Have you tested that msgs.change_licence_details_other appears? It doesn't look like it will.
I'd say decision logic like this should really be in the getData function rather than the template - so that it can be tested.

packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
```diff
@@ -209,7 +209,8 @@ const getData = async request => {
     summaryTable: getLicenseeDetailsSummaryRows(permission, countryName, request),
     uri: {
       licenceSummary: LICENCE_SUMMARY.uri
-    }
+    },
+    isLicenceForYou: permission.isLicenceForYou
```
jaucourt, line 213: Rather than passing this flag to the template, for the template to have the decision logic, the decision logic should be here, supplying the correct label.

jaucourt (changes requested):

MickStein (approved):

packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
```diff
@@ -198,19 +198,26 @@ const getLicenseeDetailsSummaryRows = (permission, countryName, request) => {
...
 
   status.fromSummary = CONTACT_SUMMARY_SEEN
   await request.cache().helpers.status.setCurrentPermission(status)
   const countryName = await countries.nameFromCode(permission.licensee.countryCode)
 
+  let changeLicenceDetails = mssgs.change_licence_details_other
+
+  if (permission.isLicenceForYou) {
+    changeLicenceDetails = mssgs.change_licence_details_you
+  }
```
jaucourt, line 213: As a rule, never use `let`, use `const`. If `const` could be one of two values, generally use the ternary operator to assign a value. If it's more than that, extract the logic to a new function. It's pretty simple logic here, so a ternary is a good fit: `const changeLicenceDetails = permission.isLicenceForYou ? mssgs.change_licence_details_you : mssgs.change_licence_details_other`

The reason for not using `let` is because `const` doesn't allow you to change the value so it makes code easier to read. It's assigned once and then you always know what it is. Whereas with let, you have to check a full code block and do a let of mental parsing to check what its value is. A fuller discussion is available here: https://medium.com/dailyjs/use-const-and-make-your-javascript-code-better-aac4f3786ca1

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/summary/contact-summary/contact-summary.njk
```diff
@@ -44,9 +44,9 @@
 
             {{ csrf() }}
         </form>
-
+
```
irisfaraway, line 47: Nitpicky but can we just get rid of this whitespace?
```suggestion

```

irisfaraway (changes requested):

irisfaraway (approved):

jaucourt (approved):

PR #1535, opened 2022-11-09
Refactor licence summary page

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3099

Refactor licence summary page, bringing licence summary rows out of the nunjucks template and into the getData function so that they can be tested

### DIFF
--- a/packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
+++ b/packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
@@ -1,26 +1,19 @@
-import { getFromSummary, getData } from '../route'
-import { LICENCE_SUMMARY_SEEN, CONTACT_SUMMARY_SEEN } from '../../../../constants.js'
-import {
-  DATE_OF_BIRTH,
-  DISABILITY_CONCESSION,
-  LICENCE_LENGTH,
-  LICENCE_TO_START,
-  LICENCE_TYPE,
-  NAME,
-  NEW_TRANSACTION,
-  RENEWAL_START_DATE
-} from '../../../../uri.js'
-import '../../find-permit.js'
+import { getData } from '../route'
+import { LICENCE_SUMMARY_SEEN } from '../../../../constants.js'
+import { DATE_OF_BIRTH, LICENCE_LENGTH, LICENCE_TO_START, LICENCE_TYPE, NAME, NEW_TRANSACTION } from '../../../../uri.js'
+import findPermit from '../../find-permit.js'
 import { licenceTypeDisplay } from '../../../../processors/licence-type-display.js'
 import { addLanguageCodeToUri } from '../../../../processors/uri-helper.js'
-import moment from 'moment-timezone'
+import mappingConstants from '../../../../processors/mapping-constants.js'
 
-jest.mock('../../find-permit.js')
-jest.mock('../../../../processors/licence-type-display.js')
-jest.mock('../../../../processors/date-and-time-display.js')
-jest.mock('../../../../processors/uri-helper.js')
+jest.mock('../../../../processors/licence-type-display.js', () => ({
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
```diff
@@ -50,288 +39,250 @@ jest.mock('../../../../processors/mapping-constants.js', () => ({
...
-          birthDate: '1996-01-01'
-        },
-        isRenewal: true
-      }
+describe('licence-summary > route', () => {
+  beforeEach(jest.clearAllMocks)
 
-      mockTransactionCacheGet.mockImplementationOnce(() => permission)
+  describe('sets from summary on status cache', () => {
+    it('adds LICENCE_SUMMARY_SEEN to status.fromSummary for a renewal', async () => {
```
ScottDormand96, line 168: in this describe could we call statusCacheSet = jest.fn() once and use for all tests within?

packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
```diff
@@ -50,288 +39,250 @@ jest.mock('../../../../processors/mapping-constants.js', () => ({
...
-
-  describe('getFromSummary', () => {
-    it('should return licence-summary, if it is a renewal', async () => {
-      const result = await getFromSummary(undefined, true)
-      expect(result).toBe(LICENCE_SUMMARY_SEEN)
+const getMockRequest = ({ currentPermission = getMockPermission(), statusCache = {}, statusCacheSet = () => {} } = {}) => ({
+  cache: () => ({
+    helpers: {
+      status: {
+        getCurrentPermission: () => statusCache,
```
ScottDormand96, line 48: should be async or not?

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
```diff
@@ -45,6 +134,33 @@ export const checkNavigation = permission => {
...
+  rows.push(rg.generateStandardRow('licence_summary_name', `${licensee.firstName} ${licensee.lastName}`, NAME.uri, 'change-name'))
+  if (!permission.isRenewal) {
+    const birthDateStr = moment(permission.licensee.birthDate).locale(request.locale).format('Do MMMM YYYY')
+    rows.push(rg.generateStandardRow('licence_summary_dob', birthDateStr, DATE_OF_BIRTH.uri, 'change-birth-date'))
+  }
+  rows.push(
+    rg.generateStandardRow(
+      'licence_summary_type',
+      licenceTypeDisplay(permission, request.i18n.getCatalog()),
+      LICENCE_TYPE.uri,
```
ScottDormand96, line 151: should we have language code added on here? same for when calling other EXAMPLE.uri

ScottDormand96 (changes requested):

packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
jaucourt: No - as far as possible, you shouldn't share mocks between tests

packages/gafl-webapp-service/src/pages/summary/licence-summary/route.js
jaucourt: No, I add the language code on line 33, saves repeating it through the code then

packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
jaucourt: Yes, good spot (it doesn't seem to matter with tests, but it makes the intent clearer if we match the mock to the actual behaviour

ScottDormand96 (approved):

PR #1553, opened 2022-11-24
Welsh translation - licence details

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3151

Add translations to licence details page


packages/gafl-webapp-service/src/locales/cy.json
```diff
@@ -100,11 +100,11 @@
   "address_select_link": " cyfeiriad",
   "address_select_title_other": "Dewisiwch eu cyfeiriad",
   "address_select_title_you": "Dewiswch eich cyfeiriad",
-  "age_concession": "Age concession",
-  "age_junior_concession": "Junior (13 to 16)",
+  "age_concession": "Consesiwn oed",
+  "age_junior_concession": "Iau (13 to 16)",
```
MickStein, line 104: Should the 'to' be updated or is that the same in Welsh?

PR #1578, opened 2022-12-13
Name page is not autofilled when buying second licence for self

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3173 

When a user is purchasing a second licence for self the name page should autofill with their name.


irisfaraway (commented):
This looks good to me (and yay for adding tests!) However do we need to do the same for the DOB in this PR as well, based on the discussions on Jira, or is that something we will cover elsewhere?

jaucourt (commented):
Also, I checked out your branch and tried it, it doesn't actually fix the bug. I think you need to do some more investigation on what's going wrong

packages/gafl-webapp-service/src/pages/contact/name/__tests__/update-transaction.spec.js
```diff
@@ -0,0 +1,73 @@
+import { NAME } from '../../../../uri.js'
+import updateTransaction from '../update-transaction.js'
+
+const VALID_PAYLOAD = {
+  'first-name': 'Luke',
+  'last-name': 'Skywalker'
+}
```
jaucourt, line 7: Don't use static values defined outside of a test in multiple tests. It's fine to have a function that gives you a value, but the problem with these const values is that they can change. Fresh test, fresh mocks.

packages/gafl-webapp-service/src/pages/contact/name/__tests__/update-transaction.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }))
+})
+
+describe('update-transaction', () => {
+  beforeEach(jest.clearAllMocks)
+  it('gets the payload from the name page cache', async () => {
+    const mockRequest = createRequestMock(VALID_PAYLOAD)
+    await updateTransaction(mockRequest)
+    expect(mockRequest.cache.mock.results[0].value.helpers.page.getCurrentPermission).toBeCalledWith(NAME.page)
```
jaucourt, line 43: This doesn't read very well. Rather than trying to fish out such a deep-lying value, in createRequestMock allow getCurrentPermission to be passed in, then you can set this up inside of the test, like this:
```
const samplePayload = {
   'first-name': 'Luke',
   'last-name': 'Skywalker'
 }
const getCurrentPermission = jest.fn(() => samplePayload)
const mockRequest = createRequestMock(samplePayload, getCurrentPermission)
await updateTransaction(mockRequest)
expect(getCurrentPermission).toHaveBeenCalledWith(NAME.page)
```

packages/gafl-webapp-service/src/pages/contact/name/__tests__/update-transaction.spec.js
```diff
@@ -0,0 +1,73 @@
...
+  })
+
+  it('gets the licensee', async () => {
+    const mockRequest = createRequestMock(VALID_PAYLOAD)
+    await updateTransaction(mockRequest)
+    expect(mockRequest.cache.mock.results[0].value.helpers.transaction.getCurrentPermission).toBeCalled()
```
jaucourt, line 49: I don't understand the point of this test - the preceding test checks that getCurrentPermission has been called?

packages/gafl-webapp-service/src/pages/contact/name/__tests__/update-transaction.spec.js
```diff
@@ -0,0 +1,73 @@
...
+const VALID_PAYLOAD = {
+  'first-name': 'Luke',
+  'last-name': 'Skywalker'
+}
+
+const transactionHelperMock = {
+  get: jest.fn(),
+  getCurrentPermission: jest.fn(() => ({ licensee: { firstName: null, lastName: null } })),
+  setCurrentPermission: jest.fn()
+}
```
jaucourt, line 13: Same goes for this

packages/gafl-webapp-service/src/pages/contact/name/__tests__/route.spec.js
```diff
@@ -1,23 +1,45 @@
 import { getData } from '../route'
 
-describe('name > route', () => {
-  const mockTransactionCacheGet = jest.fn()
+const getMockPermission = ({ isLicenceForYou }) => ({
```
jaucourt, line 3: If you default this to true, you can eliminate a bit of code in a couple of tests:
```
const getMockPermission = ({ isLicenceForYou } = { isLicenceForYou: true }) => ({
```

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/contact/name/__tests__/update-transaction.spec.js
```diff
@@ -0,0 +1,55 @@
...
+describe('update-transaction', () => {
+  beforeEach(jest.clearAllMocks)
+  it('gets the payload from the name page cache', async () => {
+    const samplePayload = { payload: { 'first-name': 'firstName', 'last-name': 'lastName' } }
+    const licenseeData = { licensee: { firstName: 'Anikan', lastName: 'Skywalker' } }
+    const getCurrentPermission = jest.fn(() => samplePayload)
+    const mockRequest = createRequestMock(getCurrentPermission, licenseeData)
+    await updateTransaction(mockRequest)
+    expect(getCurrentPermission).toHaveBeenCalledWith(NAME.page)
+  })
```
jaucourt, line 27: This test is more complicated than it needs to be: all you're testing is that `request.cache().helpers.page.getCurrentPermission` is called with `NAME.page`. Using the mock request creation function I've given above, you can do that in three lines:
```
const getPageData = jest.fn()
await updateTransaction(createRequestMock({ getPageData })
expect(getPageData).toHaveBeenCalledWith(NAME.page)
```

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/contact/name/__tests__/route.spec.js
```diff
@@ -1,23 +1,43 @@
...
+        firstName: 'Mynames',
+        lastName: 'Jeff'
+      })
+    )
+  })
+
+  it('should return isLicenceForYou as true, if isLicenceForYou is true on the transaction cache', async () => {
+    const result = await getData(getMockRequest())
+    expect(result.isLicenceForYou).toBeTruthy()
+  })
```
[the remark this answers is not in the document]
jaucourt: I'd approved then I saw this - you should remove this test, as it's not doing anything the previous test doesn't check.

jaucourt (changes requested):

jaucourt (approved):

PR #1579, opened 2022-12-15
Trigger specific error response codes from route

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3123

This PR adds a new route for testing specific error response codes. So going to /buy/throw-error?error=401 will return a 401 response code, etc.

The env var ERROR_PAGE_ROUTE must be set to true for it to work.


jaucourt (approved):
Great work!

PR #1580, opened 2022-12-16
Refactor redirects to automatically use correct language

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3111

This PR refactors internal redirects to use a new redirectWithLanguageCode function on the response toolkit. This function allows us to call a redirect for the correct language without having to also call addLanguageCodeToUri each time.  This approach was initially demoed in https://github.com/DEFRA/rod-licensing/pull/1514 but this PR handles the implementation of the function and rolling it out to the rest of the app.

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
@@ -157,7 +157,7 @@ describe('The page handler function', () => {
 
     await get(request, toolkit)
 
-    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(request, url)
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(url)
   })
 
   it.each([['/route/one'], ['/route/sixty-six']])('error redirects to request uri', async url => {
@@ -165,7 +165,7 @@ describe('The page handler function', () => {
     const request = getMockRequest(undefined, url)
     const toolkit = getMockToolkit()
     await error(request, toolkit, { details: [] })
-    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(request)
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalled()
   })
 
   it('sets the value of pageData with displayAnalytics true', async () => {
--- a/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
@@ -160,12 +160,12 @@ describe('The page handler function', () => {
     expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(url)
   })
 
-  it.each([['/route/one'], ['/route/sixty-six']])('error redirects to request uri', async url => {
+  it.each([['/route/one'], ['/route/sixty-six']])('error redirects with language code', async url => {
     const { error } = pageHandler('', 'view')
     const request = getMockRequest(undefined, url)
     const toolkit = getMockToolkit()
     await error(request, toolkit, { details: [] })
-    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalled()
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith()
   })
 
   it('sets the value of pageData with displayAnalytics true', async () => {
--- a/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
@@ -145,51 +145,27 @@ describe('The page handler function', () => {
     )
   })
 
-  it.each([['/go/somewhere'], ['/go/somewhere/else']])(
-    'GetDataRedirect being thrown will pass url to be decorated to addLanguageCodeToUri',
-    async redirectUri => {
-      journeyDefinition.push({ current: { page: 'view' }, backLink: '/previous/page' })
-      const getData = () => {
-        throw new GetDataRedirect(redirectUri)
-      }
-      const { get } = pageHandler(null, 'view', '/next/page', getData)
-      const toolkit = getMockToolkit()
-      const request = getMockRequest()
-
-      await get(request, toolkit)
-
-      expect(addLanguageCodeToUri).toHaveBeenCalledWith(request, redirectUri)
-    }
-  )
-
-  it('GetDataRedirect being thrown will use addLanguageCodeToUri to decorate the redirect target', async () => {
+  it('GetDataRedirect being thrown will redirectWithLanguageCode to the target', async () => {
     journeyDefinition.push({ current: { page: 'view' }, backLink: '/previous/page' })
+    const url = Symbol('/go/somewhere/else')
     const getData = () => {
-      throw new GetDataRedirect('/go/somewhere/else')
+      throw new GetDataRedirect(url)
     }
     const { get } = pageHandler(null, 'view', '/next/page', getData)
-    const returnValue = Symbol('/previous/page')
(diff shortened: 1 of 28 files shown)


jaucourt (changes requested):
Looks good, so nice to get all that extraneous code, just a couple of changes needed

packages/gafl-webapp-service/src/server.js
```diff
@@ -179,6 +180,12 @@ const init = async () => {
    */
   server.decorate('request', 'cache', cacheDecorator(sessionCookieName))
 
+  server.decorate('toolkit', 'redirectWithLanguageCode', function (request, redirect) {
+    const uriWithLanguage = addLanguageCodeToUri(request, redirect)
```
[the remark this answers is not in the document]
irisfaraway: There are still a few instances where we use `addLanguageCodeToUri` without a redirect, for example when we are determining where the "Back" link in the journey should go. Otherwise I would have just rolled it in here.

packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
```diff
@@ -145,51 +145,27 @@ describe('The page handler function', () => {
...
-  it.each([['/route/one'], ['/route/sixty-six']])('error redirects to uri decorated by addLanguageCodeToUri', async url => {
+  it.each([['/route/one'], ['/route/sixty-six']])('error redirects to request uri', async url => {
     const { error } = pageHandler('', 'view')
+    const request = getMockRequest(undefined, url)
     const toolkit = getMockToolkit()
-    addLanguageCodeToUri.mockReturnValueOnce(`Redirect to url ${url}`)
-    await error(getMockRequest(undefined, url), toolkit, { details: [] })
-    expect(toolkit.redirect).toHaveBeenCalledWith(`Redirect to url ${url}`)
+    await error(request, toolkit, { details: [] })
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(request)
```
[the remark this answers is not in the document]
irisfaraway: Good spot, will fix!

packages/gafl-webapp-service/src/server.js
```diff
@@ -179,6 +180,12 @@ const init = async () => {
    */
   server.decorate('request', 'cache', cacheDecorator(sessionCookieName))
 
+  server.decorate('toolkit', 'redirectWithLanguageCode', function (request, redirect) {
```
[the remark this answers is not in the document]
irisfaraway: Ahh good spot – I think this was a hangover from an earlier implementation outside server.js. Will sort it 👍

packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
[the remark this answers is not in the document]
irisfaraway: So the code this was testing was originally calling `addLanguageCodeToUrl` with just the `request` argument, which means that now we are just calling `redirectWithLanguageCode()` with no arguments. I've updated the test to clarify what we expect it to be doing.

packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
```diff
@@ -145,51 +145,27 @@ describe('The page handler function', () => {
...
-  it.each([['/route/one'], ['/route/sixty-six']])('error redirects to uri decorated by addLanguageCodeToUri', async url => {
+  it.each([['/route/one'], ['/route/sixty-six']])('error redirects with language code', async url => {
     const { error } = pageHandler('', 'view')
+    const request = getMockRequest(undefined, url)
     const toolkit = getMockToolkit()
-    addLanguageCodeToUri.mockReturnValueOnce(`Redirect to url ${url}`)
-    await error(getMockRequest(undefined, url), toolkit, { details: [] })
-    expect(toolkit.redirect).toHaveBeenCalledWith(`Redirect to url ${url}`)
+    await error(request, toolkit, { details: [] })
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith()
```
jaucourt, line 168: `toHaveBeenCalled` would be semantically accurate here. if you want to check it was called with no arguments (which might be a good test) then I'd suggest examining the mock method that there were no arguments passed.

packages/gafl-webapp-service/src/handlers/__tests__/renewal-start-date-validation-handler.spec.js
```diff
@@ -0,0 +1,57 @@
...
+
+describe('Renewal start date validation handler', () => {
+  const getMockRequest = (startYear, startMonth, startDay, renewedEndDate) => ({
+    cache: () => ({
+      initialize: () => ({}),
+      helpers: {
+        page: {
+          getCurrentPermission: async () => ({
+            payload: {
+              'licence-start-date-year': startYear || 1970,
```
jaucourt, line 12: Rather than using this construct, I'd suggest using default arguments would be a bit easier to read:
```
const getMockRequest = (startYear = 1970, startMonth = 1, startDat = 1, renewedEndDate = '2023-01-01T00:00:00.000Z') => ({
```

packages/gafl-webapp-service/src/handlers/__tests__/renewals-friendly-url-handler.spec.js
```diff
@@ -30,19 +30,21 @@ describe('The url handler', () => {
...
   })
 
   it.each([['B2F11U'], ['AH56F6'], ['GH330P']])('6 digit reference number exists and returns ATTRIBUTION', async licenceKey => {
     const params = {
       [QUERYSTRING_LICENCE_KEY]: licenceKey
     }
+    const request = generateRequestMock(params)
     const responseToolkit = generateResponseToolkitMock()
-    await urlHandler(generateRequestMock(params), responseToolkit)
+    await urlHandler(request, responseToolkit)
```
jaucourt, line 45: I've noticed you doing this in a couple of places. A pattern I use is when there's no further use for a mock, I define it inline. I suppose by this logic we should do 
```
await urlHandler(generateRequestMock({ [QUERYSTRING_LICENCE_KEY]: licenceKey }), generateResponseToolkitMock())
```
What do you think - do you feel it adds to the readability to break out request and params?

jaucourt (changes requested):
Sorry, just a few more small things

packages/gafl-webapp-service/src/handlers/__tests__/renewals-friendly-url-handler.spec.js
irisfaraway: Hah, yes, I initially broke those out when `request` needed to be tested as an argument but now that's no longer the case! I did consider changing it back, but personally I find it a lot more readable when the arguments are defined separately rather than inline, and for me it makes it easier to parse the tests. I'm happy to stick with whatever convention we decide though.

packages/gafl-webapp-service/src/handlers/__tests__/renewal-start-date-validation-handler.spec.js
irisfaraway: Good call – will sort this out tomorrow.

packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
irisfaraway: Yeah, I flipflopped between `toHaveBeenCalled()` and `toHaveBeenCalledWith()` with no arguments. I'll have a look at examining the method.

irisfaraway: On further testing it looks like `toHaveBeenCalledWith()` does check if `redirectWithLanguageCode` is called with no arguments, and fails if it is called with arguments. Personally I am happy with this and feel like it is more consistent with the rest of the tests, but can rewrite if you would rather we test it in a different way.

jaucourt (approved):

PR #1593, opened 2023-01-03
Notification banner

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3179

Implement notification banner - in this occurence for price changes

### DIFF
--- a/packages/gafl-webapp-service/src/pages/licence-details/licence-to-start/__tests__/route.spec.js
+++ b/packages/gafl-webapp-service/src/pages/licence-details/licence-to-start/__tests__/route.spec.js
@@ -27,8 +27,8 @@ describe('licence-to-start > route', () => {
     })
 
     it.each([
-      [true, true],
-      [false, false],
+      ['true', true],
+      ['false', false],
       [undefined, false]
     ])('SHOW_NOTIFICATION_BANNER is set to value of process.env.SHOW_NOTIFICATION_BANNER', async (notification, expectedResult) => {
       mockTransactionCacheGet.mockImplementationOnce(() => ({ isLicenceForYou: true }))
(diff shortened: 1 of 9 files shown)


packages/gafl-webapp-service/src/pages/licence-details/licence-to-start/licence-to-start.njk
```diff
@@ -58,7 +59,16 @@
     }) }}
 {% endset -%}
 
+{% block notificationContent %}
+
+    {% if SHOW_NOTIFICATION_BANNER %}
```
jaucourt, line 64: Does this work? I'm not sure you have direct access to environment variables in a nunjucks template, I think you have to make them available in the `getData` method

packages/gafl-webapp-service/src/pages/summary/licence-summary/__tests__/route.spec.js
```diff
@@ -244,6 +244,18 @@ describe('licence-summary > route', () => {
 
       expect(licenceTypeDisplay).toHaveBeenCalledWith(mockPermission, catalog)
     })
+
+    it.each([
+      [true, true],
+      [false, false],
```
jaucourt, line 250: the first element of these two arrays should be string values, not booleans, to mimic the input from an env file. The reason it works is that when you assign values to process.env it casts them to a string.

packages/gafl-webapp-service/src/pages/licence-details/licence-to-start/__tests__/route.spec.js
```diff
@@ -25,5 +25,16 @@ describe('licence-to-start > route', () => {
       const result = await getData(mockRequest)
       expect(result.isLicenceForYou).toBeFalsy()
     })
+
+    it.each([
+      [true, true],
+      [false, false],
```
jaucourt, line 31: First element in the array should be a string

jaucourt (changes requested):
Nice work, couple of small changes and we're there

jaucourt (approved):

PR #1594, opened 2023-01-04
Add new cost fields

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3183

Add the new cost fields to the Permit entity


packages/dynamics-lib/src/entities/__tests__/permit.entity.spec.js
```diff
@@ -22,6 +22,8 @@ describe('permit entity', () => {
         defra_iscountersales: true,
         defra_recurringsupported: false,
         defra_advertisedprice: 6.0,
+        defra_newadvertisedprice: 7.2,
+        defra_dateofpricechange: '2023-04-01T00:00:00Z',
```
irisfaraway, line 26: Just a thought – do we need to factor daylight savings into this? It starts on the 26th of March this year.

irisfaraway (commented):
This looks good! Just had a thought about dates...

ScottDormand96 (approved):

packages/dynamics-lib/src/entities/__tests__/permit.entity.spec.js
jaucourt: This year it shouldn't affect us as the clocks go forward on the 26th March (previous years it has been a bit of a nightmare when doing the DD licences that start on 1st April)

irisfaraway: I guess it depends if the dates are stored in BST or UTC. Wouldn't `2023-04-01T00:00:00Z` UTC be `'2023-04-01T01:00:00Z'` BST? This stuff always scrambles my brain.

I realise this is just test data so may not matter anyway :)

MickStein (approved):

PR #1598, opened 2023-01-06
Use permission start date to calculate price

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3184

Prices are now increasing, so the permission price needs to be calculated using the permission start date


packages/gafl-webapp-service/src/processors/__tests__/pricing-summary.spec.js
```diff
@@ -1,347 +1,562 @@
...
-      cost: 0,
-      concessions: true
-    },
-    msg: 'no-short'
-  },
-  [licenseTypes.salmonAndSeaTrout]: {
-    '12M': {
-      cost: 0,
-      concessions: true
+jest.mock('../filter-permits.js', () => ({
```
ScottDormand96, line 3: one comment - is it possible to create a createPermit function for example and pass in parameters that change instead repeating the creation?

ScottDormand96 (changes requested):

packages/gafl-webapp-service/src/processors/__tests__/pricing-summary.spec.js
jaucourt: Not really - if you look at each item, there's enough variation that it doesn't really make sense to have such a method. If you did, it would end up not saving you anything.

ScottDormand96 (approved):

PR #1613, opened 2023-01-25
Update transaction journal to use new price

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3187

Update transaction journal creation to use new price calculation in business rules lib
Also, a bit of tidying up of code written for previous tickets and a bug fix for the order complete page where prices that had pence were displaying without a trailing zero when they only had one decimal digit (e.g. 49.50 displaying as 49.5)

### DIFF
--- a/packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
+++ b/packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
@@ -31,11 +31,12 @@ export const getData = async request => {
   await request.cache().helpers.status.set({ [COMPLETION_STATUS.completed]: true })
   await request.cache().helpers.status.setCurrentPermission({ currentPage: ORDER_COMPLETE.page })
   const startTimeStringTitle = displayStartTime(request, permission)
+  const permissionCost = getPermissionCost(permission)
 
   return {
     startTimeStringTitle,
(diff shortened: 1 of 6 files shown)


packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
```diff
@@ -31,11 +31,12 @@ export const getData = async request => {
   await request.cache().helpers.status.set({ [COMPLETION_STATUS.completed]: true })
   await request.cache().helpers.status.setCurrentPermission({ currentPage: ORDER_COMPLETE.page })
   const startTimeStringTitle = displayStartTime(request, permission)
+  const permissionCost = getPermissionCost(permission)
```
ScottDormand96, line 34: when getPermissionCost is called elsewhere it has parameters of startDate and permit, will this work with permission?

jaucourt: Yes, because it's the permission that's in the transaction cache. The issue elsewhere is that we're making, essentially, an HTTP request with the permission by sending to the Sales API, and so we reduce the size of the payload by stripping out the permit details other than the permit id and then requesting this from the Sales API.

ScottDormand96 (approved):

MickStein (approved):

PR #1629, opened 2023-02-07
Improve test coverage

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-2924

This PR covers adding additional tests to fill the gaps in our unit test coverage.

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
@@ -206,6 +206,15 @@ describe('The page handler function', () => {
       })
     )
   })
+
+  it('the post method calls redirectWithLanguageCode correctly when completion is not a function', async () => {
+    const { post } = pageHandler('', 'view', '/next/page')
+    const toolkit = getMockToolkit()
+    const mockRequest = getMockRequest()
+    await post(mockRequest, toolkit)
+
+    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith('/next/page')
+  })
 })
 
 const getMockRequest = (setCurrentPermission = () => {}, path = '/buy/we/are/here', includeAnalytics = true) => ({
--- a/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
@@ -207,14 +207,17 @@ describe('The page handler function', () => {
     )
   })
 
-  it('the post method calls redirectWithLanguageCode correctly when completion is not a function', async () => {
-    const { post } = pageHandler('', 'view', '/next/page')
-    const toolkit = getMockToolkit()
-    const mockRequest = getMockRequest()
-    await post(mockRequest, toolkit)
+  it.each([['/next/page'], ['/another/page']])(
+    'post method called redirectWithLanguageCode with completion value when completion is a string',
+    async pageUri => {
+      const { post } = pageHandler('', 'view', pageUri)
+      const toolkit = getMockToolkit()
+      const mockRequest = getMockRequest()
+      await post(mockRequest, toolkit)
 
-    expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith('/next/page')
-  })
+      expect(toolkit.redirectWithLanguageCode).toHaveBeenCalledWith(pageUri)
+    }
+  )
 })
 
 const getMockRequest = (setCurrentPermission = () => {}, path = '/buy/we/are/here', includeAnalytics = true) => ({
--- a/packages/gafl-webapp-service/src/handlers/analytics-handler.js
+++ b/packages/gafl-webapp-service/src/handlers/analytics-handler.js
@@ -29,7 +29,11 @@ export default async (request, h) => {
   const { payload } = request
   const analytics = await request.cache().helpers.analytics.get()
 
-  if (analytics[ANALYTICS.selected] !== true) {
+  if (analytics[ANALYTICS.selected] === true) {
+    await request.cache().helpers.analytics.set({
+      [ANALYTICS.seenMessage]: true
+    })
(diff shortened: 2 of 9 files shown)


packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
```diff
@@ -206,6 +206,15 @@ describe('The page handler function', () => {
       })
     )
   })
+
+  it('the post method calls redirectWithLanguageCode correctly when completion is not a function', async () => {
```
jaucourt, line 210: I think a better name would be something like 'post method called redirectWithLanguageCode with completion value when completion is a string'. Might also be a candidate for it.each, using different completion urls (then you could replace 'completion value' with the url being used.)

packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -29,7 +29,11 @@ export default async (request, h) => {
   const { payload } = request
   const analytics = await request.cache().helpers.analytics.get()
 
-  if (analytics[ANALYTICS.selected] !== true) {
+  if (analytics[ANALYTICS.selected] === true) {
+    await request.cache().helpers.analytics.set({
+      [ANALYTICS.seenMessage]: true
+    })
+  } else {
```
jaucourt, line 36: I think you could combine this with the following `if`, given that there are no other clauses in the block. So:
```
} else if (payload.analyticsReponse === 'accept') {
```

packages/gafl-webapp-service/src/pages/contact/contact/__tests__/update-transaction.spec.js
```diff
@@ -258,5 +300,34 @@ describe('contact > update-transaction', () => {
         })
       )
     })
+
+    it('should set email, preferredMethodOfReminder and preferredMethodOfConfirmation in cache when how-contacted is none and preferredMethodOfNewsletter is email', async () => {
```
jaucourt, line 304: This is looking similar to previous tests - is there any way we could combine them into a single test using `each` to vary the arguments?

packages/gafl-webapp-service/src/pages/licence-details/licence-length/__tests__/result-function.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+    it('should return OK if status is not from summary and licenceToStart is after payment', async () => {
+      mockTransactionCacheGet.mockImplementationOnce(() => ({
+        licenceToStart: licenceToStart.AFTER_PAYMENT
+      }))
+      const result = await resultFunction(mockRequest)
+      expect(result).toBe(CommonResults.OK)
+    })
+
+    it('should return OK if status is not from summary and licenceLength is 12 months', async () => {
```
jaucourt, line 41: With this and the preceding test, is it worth explicitly setting `fromSummary: false` to match the test description?

packages/gafl-webapp-service/src/pages/licence-details/licence-length/__tests__/result-function.spec.js
```diff
@@ -0,0 +1,54 @@
+import resultFunction from '../result-function'
+import { CommonResults } from '../../../../constants.js'
+import { licenceToStart } from '../../licence-to-start/update-transaction.js'
+
+describe('licence-length > result-function', () => {
+  const mockStatusCacheGet = jest.fn(() => ({}))
+  const mockTransactionCacheGet = jest.fn(() => ({}))
```
jaucourt, line 7: This has always been a grotty pattern, please could you incorporate into the mockRequest?

packages/gafl-webapp-service/src/pages/licence-details/licence-length/__tests__/result-function.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      helpers: {
+        status: {
+          getCurrentPermission: mockStatusCacheGet
+        },
+        transaction: {
+          getCurrentPermission: mockTransactionCacheGet
+        }
+      }
+    })
+  }
```
jaucourt, line 20: Fresh test, fresh mocks - not a reused `mockRequest` please!

packages/gafl-webapp-service/src/pages/licence-details/licence-start-time/__tests__/result-function.spec.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const mockRequest = {
+    cache: () => ({
+      helpers: {
+        status: {
+          getCurrentPermission: mockStatusCacheGet
+        }
+      }
+    })
+  }
```
jaucourt, line 15: Same here - the pattern we use is to have a function to generate a mock request, not a reused `mockRequest` object

packages/gafl-webapp-service/src/routes/__tests__/error-routes.spec.js
```diff
@@ -155,6 +170,21 @@ describe('Error route handlers', () => {
         await serverError(request, mockToolkit)
         expect(consoleErrorSpy).toHaveBeenCalled()
       })
+
+      it.each([
+        [['kl', 'vu'], 'kl', 'vu'],
+        [['si', 'we'], 'we', 'si']
+      ])('sets altLang to be other item in two-item language array', async (locales, locale, altLang) => {
```
jaucourt, line 177: I think it might be worth extracting this and the clientError into a separate describe block, given the only variation is calling `clientError` or `serverError`

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/contact/contact/__tests__/update-transaction.spec.js
irisfaraway: That was a fun exercise in abstraction :) Done!

jaucourt (approved):
Thanks for making the changes!

PR #1642, opened 2023-02-15
Remove post office content from privacy page

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3269

Update the privacy policy to remove content about the Post Office


jaucourt (changes requested):
You've just removed bulletpoint 11, rather than renumbering and removing the last - can you do that, please?

PR #1651, opened 2023-02-21
Amend permission end date to use locale time

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3324

The end date calculation for a permission is going wrong after the change to BST, where it is going back an extra day. This is because the permission start date is set as midnight in the locale time, which is 11pm the day before in GMT, and the end date calculation takes the last minute of the previous date. Therefore '2023-03-31T23:00:00.000Z', which is '2023-04-01T00:00:00.000+01:00' is resulting in an end date of '2023-03-30T23:59:00.000Z'. It should actually be
'2023-03-31T22:59:00.000Z', or '2023-03-31T23:59:00.000+01:00'.

### DIFF
--- a/packages/sales-api-service/src/services/permissions.service.js
+++ b/packages/sales-api-service/src/services/permissions.service.js
@@ -1,5 +1,5 @@
 import { Permission, Permit } from '@defra-fish/dynamics-lib'
-import { isJunior, isSenior } from '@defra-fish/business-rules-lib'
+import { isJunior, isSenior, SERVICE_LOCAL_TIME } from '@defra-fish/business-rules-lib'
 import { getGlobalOptionSetValue, getReferenceDataForEntityAndId } from './reference-data.service.js'
 import { redis } from './ioredis.service.js'
 import moment from 'moment-timezone'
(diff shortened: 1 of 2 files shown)


packages/sales-api-service/src/services/permissions.service.js
```diff
@@ -63,7 +63,7 @@ export const calculateEndDateMoment = async ({ permitId, startDate }) => {
   const permit = await getReferenceDataForEntityAndId(Permit, permitId)
   const duration = moment.duration(`P${permit.durationMagnitude}${permit.durationDesignator.description}`)
   if (permit.durationMagnitude === 12 && permit.durationDesignator.description === 'M') {
-    return moment(startDate).add(duration).subtract(1, 'day').tz('Europe/London').endOf('day')
+    return moment(startDate).tz(SERVICE_LOCAL_TIME).add(duration).subtract(1, 'day').tz(SERVICE_LOCAL_TIME).endOf('day')
```
jaucourt, line 66: I realise there's no test for the bug here, but the bug doesn't manifest when running tests - moment exhibits buggy behaviour of its' own (see https://github.com/moment/moment/issues/4743) that masks the bug that manifests once the web app is running. So all I can do is fix the bug :/

irisfaraway (approved):
LGTM. Annoying about the testing bug!

PR #1652, opened 2023-02-23
Add new field for short-term licence confirmation preferences

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3170

As a customer purchasing a short term licence I want my preferences for short term licence confirmation recorded separately to 12 month licence confirmations to avoid them being overwritten.


packages/sales-api-service/src/services/contacts.service.js
```diff
@@ -28,24 +28,37 @@ const debug = db('sales:transformers')
...
+    contact.preferredMethodOfConfirmation = await getGlobalOptionSetValue(
+      Contact.definition.mappings.preferredMethodOfConfirmation.ref,
+      contactInCRM.preferredMethodOfConfirmation.description
+    )
+  } else {
+    contact.preferredMethodOfConfirmation = await getGlobalOptionSetValue(
+      Contact.definition.mappings.shortTermPreferredMethodOfConfirmation.ref,
+      preferredMethodOfConfirmation
+    )
+  }
```
jaucourt, line 60: I'm not sure your logic for assigning the short term method of confirmation is right. You are _always_ setting it to preferredMethodOfConfirmation, whether or not it's a short term or long term licence.
Then if it's a short term licence and the contact is in the CRM, you set the long term method of confirmation to be whatever it already is. You're essentially doing this:
```
contact.preferredMethodOfConfirmation = contact.preferredMethodOfConfirmation
```
I think you need to have a decision based on whether it's an existing contact or not, and _then_ set the values, perhaps something like this:
```
const contactAlreadyExists = (contactInCRM === undefined) // n.b. check what findContactInCRM returns if there isn't one, it might not be undefined
if (contactAlreadyExists) {
  if (shortTermLicence) {
    contact.shortTermPreferredMethodOfConfirmation = preferredMethodOfConfirmation
  } else {
    contact.preferredMethodOfConfirmation = preferredMethodOfConfirmation
  }
} else {
    contact.shortTermPreferredMethodOfConfirmation = preferredMethodOfConfirmation
    contact.preferredMethodOfConfirmation = preferredMethodOfConfirmation
}
```
It might actually be that this decision logic is unnecessary, and you can actually do this:
```
  if (shortTermLicence) {
    contact.shortTermPreferredMethodOfConfirmation = preferredMethodOfConfirmation
  } else {
    contact.preferredMethodOfConfirmation = preferredMethodOfConfirmation
  }
```
You'll have to give it a try. If the latter is the case, then you don't need the `contactAlreadyExists` variable in the first block.

packages/sales-api-service/src/services/transactions/process-transaction-queue.js
```diff
@@ -54,7 +54,7 @@ export async function processQueue ({ id }) {
     isRenewal,
     isLicenceForYou
   } of transactionRecord.permissions) {
-    const contact = await resolveContactPayload(licensee)
+    const contact = await resolveContactPayload(transactionRecord.permissions[0], licensee)
```
jaucourt, line 57: You're introducing technical debt here. Have you created a ticket flagging the bug you're introducing for multibuy?

packages/sales-api-service/src/services/__tests__/contacts.service.spec.js
```diff
@@ -27,33 +28,55 @@ describe('contacts service', () => {
...
+    })
+
+    it('licence is 12 months and contact exists in CRM', async () => {
+      const mockPayload = mockContactPayload()
+      const mockPermission = {
+        licenceLength: '12M'
+      }
+      const contact = await resolveContactPayload(mockPermission, mockPayload)
+      expect(contact.preferredMethodOfConfirmation.description).toEqual(contact.shortTermPreferredMethodOfConfirmation.description)
+    })
```
jaucourt, line 79: You have several situations, I'm not sure you're covering all of them:
1) Contact exists in CRM and is buying a short term licence. Their short term confirmation preference should be updated _without changing the long term confirmation preference_
2) Contact exists in CRM and is buying a long term licence. Their long term confirmation preference should be updated _without changing the short term confirmation preference_
3) Contact doesn't exist in CRM and is buying a short term licence. Their short term confirmation preference should be updated
4) Contact doesn't exist in CRM and is buying a long term licence. Their long term confirmation preference should be updated

The additional clause introduced in the first two cases may require four tests (so six in total). An ideal use case for `it.each`, I'd say.

jaucourt (changes requested):
There're a few changes here, but remember to add the CRM changes that're needed for this change to the release notes page (if this doesn't exist, get Joe to create one and then add them)

packages/sales-api-service/src/services/transactions/process-transaction-queue.js
ScottDormand96: https://eaflood.atlassian.net/browse/IWTF-3428

packages/sales-api-service/src/services/contacts.service.js
ScottDormand96: think was agreed if short term licence it would be set regardless, so if long term licence is bought it would overwrite the short term confirmation as default

jaucourt: I'm still not convinced by the logic you've put in.
The AC says: "If the permission is for a 12 month licence, and the method of confirmation differs from the Preferred Method Of Confirmation on the Contact ID, then the latter will be update to match the Permissions confirmation method."

We don't need to check whether the method of confirmation differs, that'd be unnecessary complication, but we need to set the (long term) preferred method of confirmation if it's a long term licence or the contact doesn't exist in the CRM

So instead of the if...else, you'd do:
```
if (permission.licenceLength === '12M' || contactInCRM === undefined) {
    contact.preferredMethodOfConfirmation = await getGlobalOptionSetValue(
       Contact.definition.mappings.shortTermPreferredMethodOfConfirmation.ref,
       preferredMethodOfConfirmation
     )
}
```
Let's have a chat about this in any case

jaucourt (commented):
Let's have a discussion about the point I've brought up

jaucourt (approved):

PR #1657, opened 2023-03-09
Refactor how we determine whether a licence can be posted

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-1970

This PR covers refactoring licence-type-display.js to change how we check whether or not a licence can be posted.

### DIFF
--- a/packages/gafl-webapp-service/src/pages/licence-details/licence-length/update-transaction.js
+++ b/packages/gafl-webapp-service/src/pages/licence-details/licence-length/update-transaction.js
@@ -4,7 +4,7 @@ import * as concessionHelper from '../../../processors/concession-helper.js'
 import moment from 'moment-timezone'
 import { cacheDateFormat } from '../../../processors/date-and-time-display.js'
 import { SERVICE_LOCAL_TIME } from '@defra-fish/business-rules-lib'
-import { isPhysical } from '../../../processors/licence-type-display.js'
+import { isPhysicalOld } from '../../../processors/licence-type-display.js'
 import { licenceToStart } from '../licence-to-start/update-transaction.js'
 
 /**
@@ -16,7 +16,7 @@ const checkContactDetails = permission => {
   const preferredMethodOfReminder = permission?.licensee?.preferredMethodOfReminder
 
   const physicalContactNone =
-    isPhysical(permission) &&
+    isPhysicalOld(permission) &&
     (preferredMethodOfConfirmation === mappings.HOW_CONTACTED.none || preferredMethodOfReminder === mappings.HOW_CONTACTED.none)
 
   if (physicalContactNone) {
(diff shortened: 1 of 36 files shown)


packages/gafl-webapp-service/src/pages/licence-details/licence-length/update-transaction.js
```diff
@@ -86,6 +87,8 @@ export default async request => {
   const { payload } = await request.cache().helpers.page.getCurrentPermission(LICENCE_LENGTH.page)
   const permission = await request.cache().helpers.transaction.getCurrentPermission()
   permission.licenceLength = payload['licence-length']
+
+  await findPermit(permission, request)
```
jaucourt, line 91: Not your fault but the passing in of variables to be modified is a bit grotty; I think we should have a tech debt ticket to change this

packages/gafl-webapp-service/src/processors/find-permit.js
```diff
@@ -1,15 +1,15 @@
...
+export const findPermit = async (permission, request) => {
   /*
    * To stop repeated reads of the API with users repeatably refreshing the page, the transaction cache stores
    * a hash of itself. If the transaction cache has not changed the permit is not recalculated.
    *
    * The section of the transaction cache subject to the hashing algorithm excludes
    * name, address, or anything not effecting permit filter
    */
-  const hashOperand = (({ hash, permit, licensee, ...p }) => p)(permission)
+  const hashOperand = (({ _hash, _permit, _licensee, ...p }) => p)(permission)
```
jaucourt, line 12: By doing this, you've made it so that the hash, permit and licensee aren't stripped from the object. If you want to maintain the _ prefix notation for unused variables, you'd have to do:
```
const hashOperand = (({ hash: _hash, permit: _permit, licensee: _licensee, ...p }) => p)(permission)
```
The original version is a bit more concise, although potentially less clear. I'll leave it up to you which course to take, I think it can be argued either way.
Have a look here: https://codepen.io/jaucourt/pen/poKGmzY?editors=0012

jaucourt (changes requested):
Great bit of work here Iris, you've been really pragmatic with your approach, improving tests without going too far beyond the scope of the ticket. Just a small amendment to make and a follow-on ticket to create, please.

packages/gafl-webapp-service/src/pages/licence-details/licence-length/update-transaction.js
irisfaraway: Logged as IWTF-3375 👍

packages/gafl-webapp-service/src/processors/find-permit.js
irisfaraway: Oops, my bad! Thanks for clarifying. I've opted for longer but more readable as clearly I missed what was going on the first time :)

jaucourt (approved):

PR #1679, opened 2023-04-05
Remove empty licences from basket

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3272

Empty licences in basket throwing error, remove empty licence

### DIFF
--- a/packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
+++ b/packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
@@ -1,19 +1,8 @@
 import { ADD_LICENCE, REMOVE_LICENCE } from '../../../../uri.js'
 import updateTransaction from '../update-transaction.js'
+import { createMockRequest } from '../../../../__mocks__/request.js'
 
 const getTransactionPermissionOne = () => ({
-  licensee: {
-    firstName: 'Turanga',
-    lastName: 'Leela'
-  },
-  licenceType: 'trout-and-coarse',
-  numberOfRods: '2',
-  licenceToStart: 'after-payment',
-  licenceLength: '12M',
-  permit: { cost: 12 }
-})
-
-const getTransactionPermissionTwo = () => ({
   licensee: {
     firstName: 'Turanga',
     lastName: 'Leela'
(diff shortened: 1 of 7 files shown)


packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
```diff
@@ -36,13 +37,71 @@ const getPermissionThree = () => ({
...
+  transactionPermission,
+  setTransaction = jest.fn(),
+  page,
+  setPage = jest.fn(),
+  status,
+  setStatus = jest.fn(),
+  setCurrentStatusPermission = jest.fn(),
+  addressLookup,
+  addressLookupPermission,
+  setAddressLookup = jest.fn()
```
jaucourt, line 83: You shouldn't create a function with so many parameters - once you get above three or four, it's far preferable to have a 'config object' so that if a calling function only has to provide a single parameter, but it's in the 11th position, you don't have to provide the previous ten parameters.

Also, do you _really_ need to make _all_ of these configurable? As far as I can see, for `transactionPermission`, you always send the result of `getTransactionPermissionThree`. So get rid of `getTransactionPermissionThree`, get rid of the parameter and just use that object where you've got `transactionPermission`

packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
```diff
@@ -53,24 +112,243 @@ describe('remove-licence > update transaction', () => {
...
+    permissions: [getPagePermissionOne(), getPagePermissionTwo(), getPagePermissionThree()]
+  }
+
+  const status = {
+    permissions: [getStatusPermissionOne(), getStatusPermissionTwo(), getStatusPermissionThree()]
+  }
+
+  const addressLookup = {
+    permissions: [getAddressLookupPermissionOne(), getAddressLookupPermissionTwo(), getAddressLookupPermissionThree()]
+  }
```
jaucourt, line 129: I've mentioned I don't know how many times not to do this...

"Fresh test, fresh mocks"

jaucourt (changes requested):
A massive well done on getting to the bottom of this.

However - your tests are far too complex, and you don't seem to have followed most of what I've written in https://eaflood.atlassian.net/wiki/spaces/IWTF/pages/4295622851/Common+Code+Review+Issues. I've put in one or two points, but if you could review your code against these common occurring issues, paying particular attention to my point about "Keep your mocks as simple as possible. If something doesn’t have to vary, don’t add a parameter for it. Use default values that cover as many tests as possible, only varying when you have to."

jaucourt (approved):

PR #1692, opened 2023-04-17
Display pre payment errors

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3299

Throw an error to be able to display pre payment error


jaucourt (commented):
I don't understand this PR - you've deleted a ton of error handlers that we need, I suspect there's been a misunderstanding from our conversation yesterday

PR #1693, opened 2023-04-20
Fix no tags after v

jaucourt:
There is a long standing problem where if any tags are created starting with v, w, x, y or z they'll cause the versioning script to break, as it relies on alphabetical order to grab the last version. For example, if a tag is created 'youreIt', this comes after all standard version tags (v1.32.0-rc.1) in alphabetical order. To fix this we can pipe the versions through a grep that'll elimate all non-version tagS


irisfaraway (approved):
Nice!

PR #1695, opened 2023-04-21
Update GA secret property env var name

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3389

Primary property needs to be named correctly, rather than xgov name it has at present.


docker/env/gafl_webapp.env.example
```diff
@@ -26,8 +26,7 @@ SALES_API_URL=http://host.docker.internal:4000
 DEBUG=webapp:*
 
 # Tracking
-#ANALYTICS_PRIMARY_PROPERTY=UA-124887477-9
-#ANALYTICS_XGOV_PROPERTY=UA-124887477-10
+#ANALYTICS_PRIMARY_PROPERTY=G-XXXXXXXX
```
jaucourt, line 29: You've only updated ANALYTICS_PRIMARY_PROPERTY, but you need to add ANALYTICS_PROPERTY_API as well, don't you?

jaucourt (approved):

PR #1700, opened 2023-04-25
Remove obsolete analytics attribution data

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3451

Analytics attribution code is no longer used in hapi-gapi and therefore needs to be removed.


packages/gafl-webapp-service/src/uri.js
jaucourt, line 1: You also need to remove RENEWAL_LICENCE, as you've removed the route in misc-routes

jaucourt (changes requested):

jaucourt (approved):

PR #1717, opened 2023-05-16
Duplicate licences when one is deleted both are removed

MickStein:

User should be able to remove a licence and the other remain.


MickStein:
This has been resolved by Scott.

PR #1723, opened 2023-05-22
Refactor findPermit Function

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3375

This function takes a permission and request as arguments but then modifies both of them. We should not be modifying the variables.


packages/gafl-webapp-service/src/processors/find-permit.js
```diff
@@ -1,8 +1,34 @@
...
+
+  const hash = crypto.createHash('sha256')
+  hash.update(JSON.stringify(hashOperand))
+  const updatedHash = hash.digest('hex')
+  const updatedRequest = await request.cache().helpers.transaction.setCurrentPermission({
+    ...updatedPermission,
+    hash: updatedHash
+  })
+  return updatedRequest
+}
```
jaucourt, line 30: What you're doing here is barely changed from what happened originally. What addHashAndPermit is doing is still complex and wrong.
There should be a function that will retrieve a permit for the permission and return it to the caller.
Then there should be another function that will create an updated hash for a permission and return it to the caller.
Neither function should be saving anything. The name doesn't suggest that this is what happens, and in any case this should be done by the calling function so that it maintains control of its data.

jaucourt (changes requested):

packages/gafl-webapp-service/src/processors/find-permit.js
```diff
@@ -1,45 +1,43 @@
...
-  const addHashAndPermit = async () => {
-    const permit = await filterPermits(permission)
-    permission.permit = permit
-    if (permit) {
-      if (!permit.newCostStartDate || !permit.newCost) {
-        debug('permit missing new cost details', permission)
-      }
-    } else {
-      debug("permit wasn't retrieved", permission)
+export const retrievePermit = async permission => {
```
jaucourt, line 7: I don't understand why you're exporting this - it's never used outside of this file

packages/gafl-webapp-service/src/processors/find-permit.js
```diff
@@ -1,45 +1,43 @@
...
+    const currentHash = updateHash(permitPermission)
+    if (currentHash !== permitPermission.hash) {
+      permitPermission.permit = await retrievePermit(permitPermission)
+      permitPermission.hash = currentHash
     } else {
       debug("permit data present and doesn't need updating")
     }
   }
+
+  return permitPermission
```
jaucourt, line 42: This is closer, but still not what I asked for. If you think about the name 'findPermit' and don't look at the code, what would you expect it to do? I know that I'd expect it to find and return a permit. I wouldn't expect it to update my passed-in permission with the permit details. I _certainly_ wouldn't expect it to update the hash of the passed-in permission.

So you need to modify this function in light of this (I think that retrievePermit is probably doing what findPermit should be doing) and then modify the calling code to take account of the revised behaviour of findPermit.

PR #1725, opened 2023-05-24
Remove inset text about concession age

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3489

Remove inset text on what is your date of birth? page


irisfaraway (changes requested):
Can we remove the `dob_inset` text from the language files as well so it's not lingering around?

irisfaraway (approved):

PR #1729, opened 2023-05-26
Order complete display new price

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3509

Update how we get price  by using transaction instead of permission to display correctly


packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
```diff
@@ -36,8 +37,8 @@ export const getData = async request => {
   return {
     startTimeStringTitle,
     isSalmonLicence: permission.licenceType === mappings.LICENCE_TYPE['salmon-and-sea-trout'],
-    permissionCost: displayPermissionPrice(permission, request.i18n.getCatalog()),
-    permissionIsFree: getPermissionCost(permission) === 0,
+    permissionCost: displayPermissionPrice(permission, request.i18n.getCatalog(), transaction.payment.created_date),
```
irisfaraway, line 40: Do we need some tests to make sure this works if the transaction does not have a payment (eg for junior licences)?

irisfaraway (commented):
Some questions about testing

packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
jaucourt: I was just about to ask this!

irisfaraway (approved):
LGTM! :shipit:

PR #1732, opened 2023-05-31
Duplicate licence in basket

ScottDormand96:

Error when duplicate licences in basket

### DIFF
--- a/packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
+++ b/packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
@@ -96,18 +96,21 @@ describe('remove-licence > update transaction', () => {
     expect(mockRequest.cache().helpers.status.setCurrentPermission()).toEqual(expect.objectContaining(setPermission))
   })
 
-  it('transaction.set is being called with a permission removed with no duplciates', async () => {
+  it('transaction cache is updated after permission is deleted where there are no duplicates', async () => {
     hasDuplicates.mockReturnValueOnce(false)
     const transaction = { permissions: [getTransactionPermissionOne(), getTransactionPermissionTwo(), getTransactionPermissionThree()] }
-    const mockRequest = createMockRequest({ cache: { transaction: transaction } })
+    const mockRequest = createMockRequest({ cache: { transaction } })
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
```diff
@@ -85,13 +96,22 @@ describe('remove-licence > update transaction', () => {
...
-  it('transaction.set is being called with a permission removed', async () => {
+  it('transaction.set is being called with a permission removed with no duplciates', async () => {
+    hasDuplicates.mockReturnValueOnce(false)
     const transaction = { permissions: [getTransactionPermissionOne(), getTransactionPermissionTwo(), getTransactionPermissionThree()] }
     const mockRequest = createMockRequest({ cache: { transaction: transaction } })
     await updateTransaction(mockRequest)
     expect(mockRequest.cache().helpers.transaction.set()).toEqual(expect.objectContaining(transaction))
   })
 
+  it('transaction.set is being called with a permission removed with duplciates', async () => {
```
jaucourt, line 107: I think you need to improve this test description. It should read 'when removing one of several identical permissions, only one of them should be removed and the others should remain'.

I think you also need to beef this test up into an it.each, where you try different cases:
1. The case you're testing at the moment with _just_ two identical permissions
2. Several permissions, all different apart from two identical permissions that aren't consecutively placed
3. Several permissions, all different apart from three identical permissions that aren't consecutively placed

packages/gafl-webapp-service/src/pages/multibuy/remove-licence/__tests__/update-transaction.spec.js
```diff
@@ -85,13 +96,22 @@ describe('remove-licence > update transaction', () => {
     expect(mockRequest.cache().helpers.status.setCurrentPermission()).toEqual(expect.objectContaining(setPermission))
   })
 
-  it('transaction.set is being called with a permission removed', async () => {
+  it('transaction.set is being called with a permission removed with no duplciates', async () => {
```
jaucourt, line 99: You need to alter this description to be a bit more 'plain English', something like 'A permission is removed on request when there are no duplicate permissions'

jaucourt (changes requested):

PR #1735, opened 2023-06-07
Allow to delete all licences

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3272

Update update transaction to only setcurrrentpermission if not null value


jaucourt (approved):
Lovely, simple change now. Good work!

PR #1736, opened 2023-06-12
Allow to delete duplicate licence

ScottDormand96:

Error when duplicate licences in basket


packages/gafl-webapp-service/src/pages/multibuy/remove-licence/update-transaction.js
```diff
@@ -1,3 +1,5 @@
+import { hasDuplicates } from '../../../processors/multibuy-processor.js'
```
jaucourt, line 1: Really you should be mocking this in your tests and faking the return value. It's ok to fake it as what the value would have been anyway.

jaucourt (changes requested):

PR #1738, opened 2023-06-21
Fix double count of page views from analytics banner

MickStein:
https://eaflood.atlassian.net/browse/IWTF-3566

When users accept tracking and hide banner the page tracks each refresh page view.

### DIFF
--- a/packages/gafl-webapp-service/src/__tests__/plugin.spec.js
+++ b/packages/gafl-webapp-service/src/__tests__/plugin.spec.js
@@ -51,7 +51,7 @@ describe('plugins', () => {
     )
   })
 
-  describe('initialiseHapiGapiPlugin', () => {
+  describe('trackAnalytics', () => {
     const generateRequestMock = (analytics = {}) => ({
       cache: jest.fn(() => ({
         hasSession: () => true,
@@ -71,7 +71,7 @@ describe('plugins', () => {
     it.each([
       [true, true],
       [false, false]
-    ])('trackAnalytics to be set to value of checkAnalytics', async (tracking, expectedResult) => {
+    ])('when acceptedTracking property equals %s, trackAnalytics should return %s', async (tracking, expectedResult) => {
       const analytics = {
         [ANALYTICS.acceptTracking]: tracking
       }
--- a/packages/gafl-webapp-service/src/__tests__/plugin.spec.js
+++ b/packages/gafl-webapp-service/src/__tests__/plugin.spec.js
@@ -1,13 +1,13 @@
 import { getPlugins } from '../plugins'
 import { ANALYTICS } from '../constants.js'
-import { checkAnalytics, getAnalyticsSessionId, skipPage } from '../handlers/analytics-handler.js'
+import { checkAnalytics, getAnalyticsSessionId, checkPage } from '../handlers/analytics-handler.js'
 import db from 'debug'
 
 jest.mock('../constants', () => ({
   ANALYTICS: {
     acceptTracking: 'accept',
     seenMessage: 'seen',
-    skipPage: 'skip'
+    skipPage: 'skip-page'
   },
   CommonResults: {
     OK: 'ok'
@@ -18,7 +18,7 @@ jest.mock('../constants', () => ({
 }))
 
 jest.mock('../handlers/analytics-handler.js', () => ({
-  skipPage: jest.fn(() => false),
+  checkPage: jest.fn(() => false),
   checkAnalytics: jest.fn(() => true),
   getAnalyticsSessionId: jest.fn(() => '123')
 }))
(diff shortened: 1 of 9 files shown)


packages/gafl-webapp-service/src/handlers/__tests__/page-handler.spec.js
```diff
@@ -12,7 +12,9 @@ jest.mock('../../constants', () => ({
   ANALYTICS: {
     selected: 'selected',
     acceptTracking: 'accepted-tracking',
-    seenMessage: 'seen-message'
+    seenMessage: 'seen-message',
+    skipPage: 'skip-page',
+    pageSkipped: 'page-skipped'
```
jaucourt, line 17: if you're mocking these, they should be different to the values in constants, otherwise there's a risk of stuff being tightly bound

packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -6,17 +6,27 @@ import { ANALYTICS } from '../constants.js'
...
-    if (analytics && analytics[ANALYTICS.acceptTracking] === true) {
+    if (analytics[ANALYTICS.acceptTracking] === true && pageSkip !== true) {
       return true
     }
   } catch {}
 
   return false
 }
 
+export const checkPage = async request => {
```
jaucourt, line 20: This needs a better name than 'checkPage' - I'd say the same is true for 'checkAnalytics' too. Look at exactly what each function is checking and base the name on that

packages/gafl-webapp-service/src/handlers/__tests__/check-analytics-handler.spec.js
```diff
@@ -1,10 +1,13 @@
...
-import { checkAnalytics, getAnalyticsSessionId } from '../analytics-handler.js'
+import { checkAnalytics, getAnalyticsSessionId, checkPage } from '../analytics-handler.js'
 
 jest.mock('../../constants', () => ({
   ANALYTICS: {
-    acceptTracking: 'accepted-tracking'
+    selected: 'selected',
+    acceptTracking: 'accepted-tracking',
+    seenMessage: 'seen-message',
+    skipPage: 'skip-page'
```
jaucourt, line 10: these should differ from what's in constants.js

packages/gafl-webapp-service/src/handlers/__tests__/check-analytics-handler.spec.js
```diff
@@ -13,20 +16,44 @@ describe('checkAnalytics', () => {
...
+  ])('when analytics skipPage property is %s, skipPage returns %s', async (skipPage, expected) => {
+    const analytics = { [ANALYTICS.skipPage]: skipPage }
+    const result = await checkPage(generateRequestMock(analytics))
+
+    expect(result).toBe(expected)
   })
 
   it('empty session cache returns false', async () => {
-    const result = await checkAnalytics(
+    const result = await checkPage(
```
jaucourt, line 56: Because of this change, you're no longer checking the try...catch in `checkAnalytics`

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -25,6 +25,23 @@ const displayAnalytics = request => {
...
+  const analytics = await request.cache().helpers.analytics.get()
+
+  if (
+    analytics &&
+    analytics[ANALYTICS.pageSkipped] !== true &&
+    analytics[ANALYTICS.seenMessage] &&
+    analytics[ANALYTICS.skipPage] !== true
+  ) {
+    await request.cache().helpers.analytics.set({ [ANALYTICS.skipPage]: true, [ANALYTICS.pageSkipped]: true })
+    return true
```
jaucourt, line 38: I don't think you're testing the return value here (or underneath). Try removing them and see if your tests all still pass. If they do, you need to add tests for the return value.

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -116,6 +133,8 @@ export default (path, view, completion, getData) => ({
 
     pageData.displayAnalytics = displayAnalytics(request)
 
+    await skipPage(request)
```
jaucourt, line 136: I don't think you're testing for this. Do any tests fail when you remove it?

Also, I don't think `skipPage` is a specific enough name - there's nothing to indicate that it's about analytics. Anyone reading this will wonder what skipPage is doing. `omitPageFromAnalytics` might be a better name?

packages/gafl-webapp-service/src/__tests__/plugin.spec.js
```diff
@@ -77,17 +83,20 @@ describe('plugins', () => {
...
     it.each([
-      [true, 'session_id_example', 'Session is being tracked for: session_id_example'],
-      [false, 'testing_session_id', 'Session is not being tracked for: testing_session_id'],
-      [true, 'example_session_id', 'Session is being tracked for: example_session_id']
-    ])('debug is called with session id and set ENABLE_ANALYTICS_OPT_IN_DEBUGGING to true', async (tracking, id, expectedResult) => {
+      [true, 'session_id_example', 'Session is being tracked for: session_id_example', false],
+      [false, 'testing_session_id', 'Session is not being tracked for: testing_session_id', false],
+      [true, 'example_session_id', 'Session is being tracked for: example_session_id', false],
+      [false, 'test_session_id', 'Session is not tracking current page for: test_session_id', true]
+    ])('debug is called with session id and set ENABLE_ANALYTICS_OPT_IN_DEBUGGING to true', async (tracking, id, expectedResult, skip) => {
```
jaucourt, line 90: No variance in the test name, so it'll be the same for all four cases.

jaucourt (changes requested):
A few amendments needed, please

packages/gafl-webapp-service/src/handlers/page-handler.js
MickStein: Yes it does look like several tests fail without this function being used. 

Good point, that name is much more descriptive will use it instead thanks.

packages/gafl-webapp-service/src/handlers/__tests__/check-analytics-handler.spec.js
MickStein: I think it might be that we added a separate test for this on line 30.

packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -25,6 +25,22 @@ const displayAnalytics = request => {
...
+  const analytics = await request.cache().helpers.analytics.get()
+
+  if (
+    analytics &&
+    analytics[ANALYTICS.pageSkipped] !== true &&
+    analytics[ANALYTICS.seenMessage] &&
+    analytics[ANALYTICS.omitPageFromAnalytics] !== true
+  ) {
+    await request.cache().helpers.analytics.set({ [ANALYTICS.omitPageFromAnalytics]: true, [ANALYTICS.pageSkipped]: true })
+    return
```
jaucourt, line 38: You shouldn't be returning nothing. You need to restructure the function to have an `else` rather than doing an empty return here.

jaucourt (changes requested):
Just one last thing 🕵️‍♀️🥴

jaucourt (approved):

PR #1741, opened 2023-06-29
Over 66 content

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3537

Updating the content for when over 66 is referenced

### DIFF
--- a/packages/gafl-webapp-service/src/processors/licence-type-display.js
+++ b/packages/gafl-webapp-service/src/processors/licence-type-display.js
@@ -7,7 +7,6 @@ export const licenceTypeDisplay = (permission, mssgs) => {
   // Build the display string for the licence type
   if (concessionHelper.hasJunior(permission)) {
     typesStrArr.push(mssgs.age_junior)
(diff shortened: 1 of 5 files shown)


irisfaraway (approved):

packages/gafl-webapp-service/src/processors/licence-type-display.js
```diff
@@ -7,10 +7,8 @@ export const licenceTypeDisplay = (permission, mssgs) => {
   // Build the display string for the licence type
   if (concessionHelper.hasJunior(permission)) {
     typesStrArr.push(mssgs.age_junior)
-  } else if (concessionHelper.hasSenior(permission)) {
-    typesStrArr.push(mssgs.over_66)
+    typesStrArr.push(', ')
```
jaucourt, line 10: This should be part of the label, rather than being added as a hard-coded string

jaucourt (changes requested):
Nice work, just one small change, please

jaucourt (approved):

PR #1752, opened 2023-07-10
Update English language version of accessibility statement

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3055

This PR makes some modifications to the English language version of the accessibility statement as documented on Jira.


irisfaraway:
For whoever reviews this – there were some inconsistencies in the text so might be worth checking the comments on Jira for an explanation.

Also it may be easier to review by looking at the page itself so deploy the `iwtf-3053-accessibility-statement-updates` tag if you want to do that :)

packages/gafl-webapp-service/src/locales/en.json
```diff
@@ -1,60 +1,65 @@
 {
-  "access_statement_ability_net_link": "AbilityNet",
-  "access_statement_ability_net_1": " has advice on making your device easier to use if you have a disability.",
+  "access_statement_ability_net_link": "AbilityNet (opens in new tab)",
+  "access_statement_ability_net_1": "",
```
jaucourt, line 3: Should we just get rid of this, rather than having an empty label?

jaucourt (changes requested):

jaucourt (approved):

PR #1753, opened 2023-07-10
Update Welsh language version of accessibility statement

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3053

This PR makes some modifications to the Welsh language version of the accessibility statement as documented on Jira.


irisfaraway:
#1752 will need to be merged before this one as that includes the HTML changes

irisfaraway:
For whoever reviews this – there were some inconsistencies in the text so might be worth checking the comments on Jira for an explanation.

Also it may be easier to review by looking at the page itself, so deploy the `iwtf-3053-accessibility-statement-updates` tag if you want to do that :)

jaucourt (approved):

PR #1784, opened 2023-09-01
Fix issues with processing postal order records

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3418

This PR fixes a few issues that came up while adding postal order records via the new CRM process.

There were a few mentions of fields prefixed with new* which had since been renamed - so these needed to be updated here as well.

Once this was changed, further errors revealed that we were passing through objects with labels rather than just the label string, which is what the sales API expected. So this PR also includes tweaks to make sure we pass through just a string when that is expected.

### DIFF
--- a/packages/pocl-job/src/staging/pocl-validation-errors.js
+++ b/packages/pocl-job/src/staging/pocl-validation-errors.js
@@ -24,7 +24,7 @@ const mapRecords = records =>
             locality: record.locality,
             town: record.town,
             postcode: record.postcode,
-            country: record.country || record.countryUnvalidated,
+            country: getCountryValue(record),
             preferredMethodOfConfirmation: record.preferredMethodOfConfirmation.label,
             preferredMethodOfNewsletter: record.preferredMethodOfNewsletter.label,
             preferredMethodOfReminder: record.preferredMethodOfReminder.label,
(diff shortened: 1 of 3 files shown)


packages/pocl-job/src/staging/pocl-validation-errors.js
```diff
@@ -119,10 +133,9 @@ const createTransactions = async records => {
 const finaliseTransaction = async rec => {
   const payment = rec.record.finaliseTransactionPayload.payment
   const finaliseTransactionPayload = {
-    payment: { method: backfillPaymentMethod(payment.method, payment.newPaymentSource) },
+    payment: { method: backfillPaymentMethod(payment.method, payment.paymentSource) },
     ...rec.record.finaliseTransactionPayload
   }
-  debug('finalising transaction: %o', finaliseTransactionPayload)
```
ScottDormand96, line 125: should we be removing this debug?

ScottDormand96 (changes requested):

packages/pocl-job/src/staging/pocl-validation-errors.js
irisfaraway: Oops, got overzealous there! Fixed

ScottDormand96 (approved):

PR #1788, opened 2023-09-04
Accessibility - back link

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3672

Remove back link from main content


irisfaraway (changes requested):
@ScottDormand96 This looks good for the pages that use the standard-form, but not all of them do. If you do a search in the codebase for `govukBackLink` you'll find quite a few pages which will need to be updated separately.

PR #1791, opened 2023-09-08
Strip fragments when processing URIs

irisfaraway:

Selecting the 'skip to main content' link will append a #main-content fragment to the URL of the page. However this fragment then persists through the rest of the journey, which basically means that if you skipped once, you are forced to skip for the rest of the journey.

This PR updates how we process URIs to remove any fragments as well as adding or retaining the language code.


irisfaraway:
I have some thoughts on whether or not this is the right way to go – see Jira

irisfaraway:
Closing as I was barking up the wrong tree with this.

PR #1795, opened 2023-09-13
Refactor findPermit function

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3375

Refactor this function so that it doesn't modify parameters anymore


packages/gafl-webapp-service/src/pages/licence-details/licence-length/__tests__/update-transaction.spec.js
```diff
@@ -1,103 +1,133 @@
...
-            getCurrentPermission: () => transactionPageGet,
-            setCurrentPermission: jest.fn()
+            getCurrentPermission: () => permission,
+            setCurrentPermission: transactionSet
           }
         }
       }))
     })
 
     it('should find the permit with the available data', async () => {
```
ScottDormand96, line 41: should we test that we are calling hashPermission with permission?

jaucourt: Good spot, I'd missed that. I've added it in now

ScottDormand96 (approved):

irisfaraway (approved):

PR #1797, opened 2023-09-15
Add current path as action for forms

irisfaraway:

This change is intended to fix an accessibility issue in which the #main-content fragment, once applied by the skip link, was carried forward through the rest of the journey, forcibly skipping the user to the main content on each subsequent page.


irisfaraway:
Closing as going with a different approach (again)

PR #1798, opened 2023-09-15
Specify language in HTML tags

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3674

This change makes sure that the correct language tags are applied to the HTML depending on whether the page is in English or Welsh. It also adds attributes to the language switch links to make sure those are always tagged with the correct language.

### DIFF
--- a/packages/gafl-webapp-service/src/pages/layout/layout.njk
+++ b/packages/gafl-webapp-service/src/pages/layout/layout.njk
@@ -9,6 +9,10 @@
 {% from "button/macro.njk" import govukButton %}
 {% from "footer/macro.njk" import govukFooter %}
 
+{% if (SHOW_WELSH_CONTENT) and (_uri.queryParams.lang == 'cy') %}
+    {% set htmlLang = "cy" %}
+{% endif %}
+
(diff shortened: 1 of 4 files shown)


packages/gafl-webapp-service/src/pages/layout/layout.njk
```diff
@@ -9,6 +9,10 @@
 {% from "button/macro.njk" import govukButton %}
 {% from "footer/macro.njk" import govukFooter %}
 
+{% if (SHOW_WELSH_CONTENT) and (_uri.queryParams.lang == 'cy') %}
```
jaucourt, line 12: Would be nice to see this as a flag in the injected data, which can then be tested, rather than logic being in the template

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/layout/layout.njk
irisfaraway: @jaucourt Good shout, updated

jaucourt (approved):

PR #1799, opened 2023-09-18
Stop redirect from always preserving skip link fragment

irisfaraway:

This change fixes an accessibility issue in which the #main-content fragment, which was applied when a user selected the 'Skip to main content' link, was then applied to all subsequent pages in the journey. This is because the 302 redirects always preserved the existing fragment.

As other possible solutions like updating the action tag then broke the language param, we decided to fix this by making sure a blank fragment is applied on redirects to override #main-content.


irisfaraway:
This will be a lot clearer to review if you look at the two commits separately, as the vast majority of line changes are just updating tests.

jaucourt (commented):
Now I see what's been necessary, I'm a little less sure about this method. I'm wondering whether using `toMatch` rather than `toBe` would be a better choice than adding a hash into every test, so for example the assertation in 'loads the order compete page' would become:
```
expect(response.headers.location).toMatch(new RegExp(`^${ORDER_COMPLETE}`))
```

jaucourt (approved):

irisfaraway:
Agreed to create a new ticket to handle future test improvements: https://eaflood.atlassian.net/browse/IWTF-3759

PR #1801, opened 2023-09-25
Pricing summary refactor

jaucourt:
Pricing summary refactor

https://eaflood.atlassian.net/browse/IWTF-3322

Refactor the pricing summary which is doing a lot of work with no caching


ScottDormand96:
I think it is easier to read with the changes you've made. Aware of risk with changing code but feel testing isn't hugely time consuming to check with all scenarios is displaying correctly. Happy to go ahead with changes if content pricing for all concessions is likely stay the same

PR #1803, opened 2023-09-27
Retrieve recurring payments

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3617

Recurring Payments Job to retrieve the recurring payments due on the current day


packages/dynamics-lib/src/entities/recurring-payment.entity.js
```diff
@@ -10,16 +11,22 @@ export class RecurringPayment extends BaseEntity {
   static _definition = new EntityDefinition(() => ({
     localName: 'recurringPayment',
     dynamicsCollection: 'defra_recurringpayments',
-    defaultFilter: 'statecode eq 0',
+    defaultFilter: 'statecode eq 0 and defra_cancelleddate eq null',
```
jaucourt, line 14: I'd question whether this should be a default filter

packages/recurring-payments-job/src/__mocks__/dynamics-web-api.js
```diff
@@ -0,0 +1,3 @@
+import { configureDynamicsWebApiMock } from '../../../dynamics-lib/src/__mocks__/dynamics-web-api-mock-helper.js'
+const DynamicsWebApi = jest.genMockFromModule('dynamics-web-api')
+export default configureDynamicsWebApiMock(DynamicsWebApi)
```
jaucourt, line 3: I'm not a fan of auto mocks - they hide functionality and make things work 'magically' in an opaque way. Mocks should be declared in the same file as the tests they're serving.

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -1,3 +1,12 @@
-export async function execute () {
-  // do something
+import { getRecurringPayments } from '../../sales-api-service/src/services/recurring-payments.service.js'
```
jaucourt, line 1: You shouldn't be calling Sales API modules in this way. They should be called via the HTTP interface - you can't rely on the sales-api-service being available, as when recurring-payments is running in its own container it won't be. For an example of how to do it, have a look at the agreed-handler.js in the GAFL. You'll see there's a salesApi import there, that actually connects to the @defra-fish/connectors-lib, which in turn makes the HTTP request to the Sales API

packages/sales-api-service/src/schema/transaction.schema.js
```diff
@@ -106,9 +106,24 @@ const finaliseTransactionRequestSchemaContent = {
...
+      cancelledDate: Joi.string()
+        .isoDate()
+        .optional()
+        .description('Optional field for when recurring payment cancelled')
+        .example(new Date().toISOString()),
+      cancelledReason: buildJoiOptionSetValidator('defra_cancelledreason', 'User Cancelled'),
+      endDate: Joi.string().isoDate().required().description('End of recurring payment').example(new Date().toISOString()),
+      agreementId: Joi.string().required().description('Agreement identification number, Gov.UK Pay field').example(uuidv4()),
+      publicId: Joi.string().required().description('SHA-256 hash of id').example(uuidv4()),
+      status: Joi.number().valid(0, 1).required().description('Status field: 1 for active, 0 for not active'),
```
jaucourt, line 124: Haven't you put in an auto-filter on this field, so it will always be 1?

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -0,0 +1,54 @@
...
+    dueDates.push(date)
+  }
+
+  const recurringPayments = []
+  for (const dueDate of dueDates) {
+    const activeRecurringPayments = await findByExample(new RecurringPayment())
+    const dueRecurringPayments = await findRecurringPaymentsInDateRange(activeRecurringPayments, dueDate)
+
+    if (dueRecurringPayments.length !== 0) {
+      recurringPayments.push(dueRecurringPayments)
```
jaucourt, line 18: This is not the way you should concatenate arrays - push is for pushing a single item onto an array. Have a look at this codepen: https://codepen.io/jaucourt/pen/YzJmQWV. You'll see that when you push one array onto another, it's added as an array within an array.

One way to concatenate arrays is to use the [concat function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat). This probably isn't the most performant, though, so you need to explore how best to do it.

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -0,0 +1,54 @@
...
+
+export const retrieveActivePermission = async recurringPayments => {
+  const recurringPaymentsWithPermission = []
+  for (const recurringPayment of recurringPayments) {
+    const permission = await findById(Permission, recurringPayment.activePermission)
+    recurringPayment.activePermission = permission
+    recurringPaymentsWithPermission.push(recurringPayment)
+  }
+  return recurringPaymentsWithPermission
+}
```
jaucourt, line 33: This isn't how you should retrieve the related permission. If you look in dynamics-lib/src/queries/permission.queries.js, you'll see that it retrieves related entities for the Permission: Contact, Permit and Concessions. You should be doing it in the same way.

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -0,0 +1,54 @@
...
+    })
+
+    const recurringPaymentsWithPermission = await retrieveActivePermission(recurringPaymentsInDateRange)
+
+    return recurringPaymentsWithPermission
+  } catch (error) {
+    console.error('Error finding recurring payments:', error)
+    throw error
+  }
+}
```
jaucourt, line 54: This also shouldn't be here - it should be a filter passed into a function defined in dynamics-lib

jaucourt (changes requested):
I can see you've had a really good crack at this, but I'm afraid it's going to need significant changes. You've departed from the established patterns of doing things - I've pointed these out. We can perhaps pair up on this if you're not confident on how to proceed.

PR #1806, opened 2023-10-02
Remove erroneous fieldset and fix aria description

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3676

### DIFF
--- a/packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
+++ b/packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
@@ -41,7 +41,7 @@
 %}
 
 {% block pageContent %}
-    <div id="date-of-birth-hint" class="govuk-hint">
+    <div id="{{ describedByTag }}" class="govuk-hint">
       {{ mssgs.dob_entry_hint }}
     </div>
 
--- a/packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
+++ b/packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
@@ -3,6 +3,7 @@
 {% from "date-input/macro.njk" import govukDateInput %}
 {% from "inset-text/macro.njk" import govukInsetText %}
 {% set title = mssgs.dob_title_you if data.isLicenceForYou else mssgs.dob_title_other %}
+{% set describedByTag = "date-of-birth-hint" %}
 
 {%
     set errorMap = {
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
```diff
@@ -3,6 +3,7 @@
 {% from "date-input/macro.njk" import govukDateInput %}
 {% from "inset-text/macro.njk" import govukInsetText %}
 {% set title = mssgs.dob_title_you if data.isLicenceForYou else mssgs.dob_title_other %}
+{% set describedByTag = "date-of-birth-hint" %}
```
ScottDormand96, line 6: wondering if way to do this in route so can test?

ScottDormand96 (changes requested):

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
jaucourt: As I see this, it's the id for an element on the page which is passed to a sub-template. As such, I think we can get away with just having it in the template rather than needing to have it in the getData function. However, I do think that if we're using a variable in this way, the div with this id should use the variable rather than having the id hard-coded.

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
```diff
@@ -40,11 +41,9 @@
 %}
 
 {% block pageContent %}
-    <fieldset class="govuk-fieldset" role="group" aria-describedby="passport-issued-hint">
-      <div id="passport-issued-hint" class="govuk-hint">
-        {{ mssgs.dob_entry_hint }}
-      </div>
-    </fieldset>
+    <div id="date-of-birth-hint" class="govuk-hint">
```
jaucourt, line 44: So here, it should be `<div id="{{ describedByTag }}" class="govuk-hint">`

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/concessions/date-of-birth/date-of-birth.njk
irisfaraway: Ahh good shout, will fix

jaucourt (approved):

ScottDormand96 (approved):

PR #1808, opened 2023-10-02
Fix markup on pricing summary box

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3680

This fixes the markup on the list of prices to use the DL syntax. It also fixes a number of duplicate IDs which should have been classes.


ScottDormand96 (approved):

jaucourt (approved):
I think the scss could be slightly tidier, but it's pretty minor so I'll leave it up to you whether you implement my suggestion or not

PR #1811, opened 2023-10-17
Insert Google Tag Manager snippet in template

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3732

This PR adds the snippet for Google Tag Manager to the default layout. This snippet should only be inserted when the container ID is present.

### DIFF
--- a/packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
@@ -12,7 +12,7 @@ exports[`The page handler function sets the value of pageData with displayAnalyt
         "analyticsSelected": false,
         "backRef": null,
         "displayAnalytics": false,
-        "gtmContainerId": null,
+        "gtmContainerId": false,
         "mssgs": undefined,
         "pageLanguageSetToWelsh": false,
         "uri": Object {
@@ -42,7 +42,7 @@ exports[`The page handler function sets the value of pageData with displayAnalyt
         "analyticsSelected": false,
         "backRef": null,
         "displayAnalytics": true,
-        "gtmContainerId": null,
+        "gtmContainerId": false,
         "mssgs": undefined,
         "pageLanguageSetToWelsh": false,
         "uri": Object {
--- a/packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
+++ b/packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
@@ -12,6 +12,7 @@ exports[`The page handler function sets the value of pageData with displayAnalyt
         "analyticsSelected": false,
         "backRef": null,
         "displayAnalytics": false,
+        "gtmContainerId": null,
         "mssgs": undefined,
         "pageLanguageSetToWelsh": false,
         "uri": Object {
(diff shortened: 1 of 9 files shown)


packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
```diff
@@ -41,6 +42,7 @@ exports[`The page handler function sets the value of pageData with displayAnalyt
         "analyticsSelected": false,
         "backRef": null,
         "displayAnalytics": true,
+        "gtmContainerId": null,
```
jaucourt, line 45: Please could we set gtmContainerId to a default value in the test so that we don't end up with a null value in the snapshot test?

packages/gafl-webapp-service/src/pages/layout/layout.njk
```diff
@@ -24,6 +24,11 @@
 <meta http-equiv="Cache-Control" content="no-store" />
 <meta http-equiv="Pragma" content="no-cache">
 <meta http-equiv="Expires" content="-1">
+{% if gtmContainerId %}
+    <!-- Google Tag Manager -->
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start': new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);})(window,document,'script','dataLayer','{{gtmContainerId}}');</script>
+    <!-- End Google Tag Manager -->
+{% endif %}
```
jaucourt, line 31: Might be over-egging this, but I wonder if it would be worth putting something in the logs if the gtmContainerId isn't present? We could put something in the page console.info, as well as in the GAFL process logs

packages/gafl-webapp-service/src/pages/layout/layout.njk
```diff
@@ -170,7 +175,13 @@
    {% endif %}
 {% endblock %}
 
-{% block bodyStart %}{% endblock %}
+{% block bodyStart %}
+{% if gtmContainerId %}
+    <!-- Google Tag Manager (noscript) -->
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{gtmContainerId}}" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
```
jaucourt, line 181: Tiny point, but we don't need `visibility:hidden` if there's already `display:none`

jaucourt (changes requested):
A few small comments. Also, I think we need a ticket to remove hapi-gapi from GAFL - would you mind creating this, please?

packages/gafl-webapp-service/src/pages/layout/layout.njk
irisfaraway: This was the code Google provided so I'm wary of messing with it!

packages/gafl-webapp-service/src/handlers/__tests__/__snapshots__/page-handler.spec.js.snap
irisfaraway: Fixed – it defaults to false instead now

packages/gafl-webapp-service/src/pages/layout/layout.njk
irisfaraway: Added logging during server startup

irisfaraway:
Ticket created to remove Hapi Gapi code: https://eaflood.atlassian.net/browse/IWTF-3843

packages/gafl-webapp-service/src/pages/layout/layout.njk
jaucourt: Yeah, whoever wrote it doesn't understand CSS!

jaucourt (approved):

PR #1822, opened 2023-11-15
Banner for end of licence cards

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3748

Banner displaying information that postal licences are changing


packages/gafl-webapp-service/src/pages/contact/digital-licence/licence-fulfilment/__tests__/route.spec.js
```diff
@@ -41,5 +41,17 @@ describe('licence-fulfilment > route', () => {
       const result = await getData(getMockRequest())
       expect(result.youOrOther).toEqual(returnedValue)
     })
+
+    it.each([
+      ['true', true],
+      ['false', false],
+      [undefined, false]
+    ])('SHOW_NOTIFICATION_BANNER is set to value of process.env.SHOW_NOTIFICATION_BANNER', async (notification, expectedResult) => {
```
jaucourt, line 49: The test name will be the same for all three iterations, it should differ so that we know what's being tested. "when SHOW_NOTIFICATION_BANNER env var is %s, flag for showing notification banner should be %s".

packages/gafl-webapp-service/src/pages/contact/digital-licence/licence-fulfilment/route.js
```diff
@@ -16,7 +16,10 @@ export const getData = async request => {
     throw new GetDataRedirect(CONTACT.uri)
   }
 
-  return { youOrOther: youOrOther(permission) }
+  return {
+    SHOW_NOTIFICATION_BANNER: process.env.SHOW_NOTIFICATION_BANNER?.toLowerCase() === 'true',
```
jaucourt, line 20: convention is to use camelCase rather than all caps with underscores - please would you change it?

jaucourt (changes requested):
Just a couple of small things to change, please

jaucourt (approved):

PR #1831, opened 2023-11-24
Pricing summary duplicate id

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3681

On the ‘Which type of licence do you want?’ page, every price element in pricing summary  has the same ID of ‘pricing-summary-element-cost’.


irisfaraway:
@ScottDormand96 I'm not sure this change is necessary – the ticket talked about changing the ID to be unique but this is changing the class. Looks like we previously ditched the ID entirely in https://github.com/DEFRA/rod-licensing/pull/1808 so either we add a new unique ID or we call this ticket already done.

PR #1845, opened 2023-12-21
Recurring Payment not for BOBO

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3898

Recurring payment is not available for BOBO so should only display if buying licence for yourself


jaucourt (approved):
LGTM!

PR #1853, opened 2024-01-10
Fix wrong price in CRM

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3932

Fix the problem where, when buying a twelve month licence to start in 30 minutes, where the price hasn’t yet gone up but will have done by the time the licence starts, the current (lower) price is quoted and charged, but the higher price is recorded in the CRM.

### DIFF
--- a/packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
+++ b/packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
@@ -187,6 +187,8 @@ describe('transaction service', () => {
       mockRecord.permissions[0].isLicenceForYou = true
       AwsMock.DynamoDB.DocumentClient.__setResponse('get', { Item: mockRecord })
       await processQueue({ id: mockRecord.id })
+
+      // assert
       const persistMockFirstAgument = persist.mock.calls[0]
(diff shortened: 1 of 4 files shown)


packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
```diff
@@ -189,6 +187,8 @@ describe('transaction service', () => {
       mockRecord.permissions[0].isLicenceForYou = true
       AwsMock.DynamoDB.DocumentClient.__setResponse('get', { Item: mockRecord })
       await processQueue({ id: mockRecord.id })
+
+      // assert
```
irisfaraway, line 191: Is this necessary?

irisfaraway (commented):
Looks good, just a question about a comment

packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
jaucourt: Argh, I was showing something to Laila and obviously committed this by accident!

irisfaraway (approved):

PR #1878, opened 2024-01-22
Recurring payment terms and conditions welsh

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3828

Updating recurring payment terms and conditions page with welsh content


packages/gafl-webapp-service/src/routes/__tests__/misc-routes-handlers.spec.js
```diff
@@ -254,6 +255,26 @@ describe('guidance page handlers', () => {
         expect(addLanguageCodeToUri).toHaveBeenCalledWith(request, referrer)
       }
     )
+
+    it.each([[CONTROLLER.uri], [RECURRING_TERMS_CONDITIONS.uri]])(
+      'back button should be set too %s when referrer is %s',
```
jaucourt, line 260: Typo on 'too', should be 'to'. Also, the second `%s` won't actually be replaced, so ends up with something like 'back button should be set too controller.uri when referrer is %s' when the test runs

jaucourt (changes requested):
Couple of small things to fix

jaucourt (approved):

PR #1897, opened 2024-02-01
Confirmation page content

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3890

Remove link for postal licence with digital confirmation


jaucourt (approved):
Looks fine, but slightly confused that something could be both a postal licence and a digital licence at the same time?

PR #1904, opened 2024-02-12
Check if the fulfilment switchover date has passed when processing transactions

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3834

We don't want to create new FulfilmentRequest objects in the CRM once the switchover date has passed. This PR adds the date and also updates our transaction queue logic to factor that date in when creating entities.


irisfaraway:
Had some indecision about where the env var override for `FULFILMENT_SWITCHOVER_DATE` should happen. constants.js felt like the wrong place for it but maybe it should be happening elsewhere in the business-rules-lib. Reviewer, lend me your thoughts!

jaucourt (approved):

PR #1909, opened 2024-02-20
Implement recurring payments

ScottDormand96:
Setup recurring payments job


packages/sales-api-service/src/services/transactions/process-transaction-queue.js
```diff
@@ -68,6 +65,9 @@ export async function processQueue ({ id }) {
       isRenewal
     )
 
+    const { recurringPayment } = await processRecurringPayment(transactionRecord, contact)
+    recurringPayment && entities.push(recurringPayment)
```
jaucourt, line 69: You should be testing for the call out to processRecurringPayment, that entities has the return value of processRecurringPayment pushed on when it returns a value and that it doesn't when processRecurringPayment returns a null or undefined.

Also, please don't use logical value operators in this way - it's ok for assigning values to a variable, but it's clearer to use an if statement in this case:
```
if (recurringPayment) {
  entities.push(recurringPayment)
}
```

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -0,0 +1,27 @@
+import { executeQuery, findDueRecurringPayments, RecurringPayment } from '@defra-fish/dynamics-lib'
+
+export const getRecurringPayments = date => executeQuery(findDueRecurringPayments(date))
+
+/**
+ * Process a recurring payment instruction
+ * @param transactionRecord
+ * @returns {Promise<{recurringPayment: null, contact: null}>}
```
jaucourt, line 8: This is incorrect, there is never a contact key so you should omit it.
Also, recurringPayment should be (RecurringPayment | null), as it isn't always a null value

jaucourt (changes requested):

jaucourt (changes requested):
One last thing, plus did you create the ticket I asked about here: https://github.com/DEFRA/rod-licensing/pull/1909#discussion_r1535304716?

jaucourt (approved):

PR #1912, opened 2024-02-21
SEMVER - PATCH: Release v1.40.1

jaucourt:
Fix leap year end date

https://eaflood.atlassian.net/browse/IWTF-4006

When twelve month licences start on the leap day (29th February) of a leap year, they end a day earlier than they should because of the order that the aritmetic is done


packages/sales-api-service/src/services/transactions/__tests__/end-date.spec.js
ScottDormand96, line 1: Should we add test in check still returns correct date when not a leap year?

ScottDormand96 (changes requested):

packages/sales-api-service/src/services/transactions/__tests__/end-date.spec.js
jaucourt: I'd argue that those tests should already be there, but having checked I can't really see any so I'll add some

ScottDormand96 (approved):

PR #1922, opened 2024-03-01
Renewing a licence starts a day early

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3958

When renewing a licence, they're starting a day early at 11pm


packages/gafl-webapp-service/src/processors/renewals-write-cache.js
```diff
@@ -53,6 +62,7 @@ export const setUpCacheFromAuthenticationResult = async (request, authentication
 
   // Add in concession proofs
   const concessions = await salesApi.concessions.getAll()
+  console.log('concessions', concessions)
```
irisfaraway, line 65: You've got some stray logging over here

irisfaraway (changes requested):
Looks good! Loving the test refactor. Just a very minor thing to change and then we're good

irisfaraway (approved):

PR #1925, opened 2024-03-07
Keyboard focus inconsistency

lailien3:
https://eaflood.atlassian.net/browse/IWTF-3838

A recent accessibility assessment identified a number of issues. This ticket is to make keyboard focus consistent on Change links with the GOV.UK Design System. The issue and suggested solution is detailed on page 118-119 of the DAC assessment.

This affects keyboard only users, as this link was styled inconsistently from other links, using the browser default styling. It now uses the yellow and black, thick underline, styling used everywhere else.

### DIFF
--- a/packages/gafl-webapp-service/package-lock.json
+++ b/packages/gafl-webapp-service/package-lock.json
@@ -9,8 +9,6 @@
       "version": "1.43.0-rc.2",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
-        "@defra-fish/business-rules-lib": "1.43.0-rc.2",
-        "@defra-fish/connectors-lib": "1.43.0-rc.2",
         "@defra/hapi-gapi": "^2.0.0",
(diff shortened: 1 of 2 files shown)


jaucourt (approved):
LGTM, but don't merge until we've taken the release cut

packages/gafl-webapp-service/package-lock.json
```diff
@@ -9,8 +9,6 @@
       "version": "1.43.0-rc.2",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
-        "@defra-fish/business-rules-lib": "1.43.0-rc.2",
-        "@defra-fish/connectors-lib": "1.43.0-rc.2",
```
jaucourt, line 13: I don't think these should have been deleted - especially not with them still being in package.lock...

jaucourt (changes requested):

jaucourt (approved):

PR #1932, opened 2024-03-14
NCCC users not see RCP option

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3910

add flag check that valid for recurring pay if not a telesales

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
@@ -28,11 +28,13 @@ describe('recurringPayReminderDisplay', () => {
 
 describe('validForRecurringPayment', () => {
   it.each([
-    [true, '12M', true, true],
-    [false, '8D', true, true],
-    [false, '12M', false, true],
-    [false, '12M', true, false]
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
```diff
@@ -28,14 +28,19 @@ describe('recurringPayReminderDisplay', () => {
...
-    process.env.SHOW_RECURRING_PAYMENTS = recurring
-    const permission = getPermission({ licenceFor, length })
-    const result = validForRecurringPayment(permission)
-    expect(result).toEqual(expected)
-  })
+    [true, '12M', true, true, undefined],
+    [false, '8D', true, true, undefined],
+    [false, '12M', false, true, undefined],
+    [false, '12M', true, false, undefined],
+    [false, '12M', true, true, 'telesales']
```
jaucourt, line 35: Tiny point, but you're not testing the journey being a different string, only `undefined`. Put in a test for that and I think we're there.

jaucourt (changes requested):

jaucourt (approved):

PR #1944, opened 2024-04-10
Fixing privacy title discrepancy

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4062

A bug was discovered on the privacy policy page. The Welsh page title matched the H1 title in English, but was different in Welsh. The page titles should match the H1 titles.


irisfaraway (changes requested):
Text changes look good to me. However you don't need to manually include the service name and the GOV.UK bit in the text here as this means we would be duplicating it across every title – so if we ever needed to change it, we'd have to change it across a bunch of different places.

Instead we define it in the language files once (as `header_service_name_title` and `gov_uk`) and then re-use it as we need to. So you'll want to update the privacy policy Nunjucks file to do this instead. Have a peek in some of the other templates and standard-form.njk for how this is done.

lailien3:
I have made changes using the macro, but have done it on a different branch [bug/iwtf-4062-fixing-privacy-title-discrepancies]

PR #1945, opened 2024-04-11
https://eaflood.atlassian.net/browse/IWTF-4062

lailien3:
A bug was discovered on the privacy policy page. The Welsh page title matched the H1 title in English, but was different in Welsh. The page titles should match the H1 titles.


irisfaraway (commented):
PR looks good to me! However before merging I think it's worth confirming that we definitely want to be changing the Welsh title of the service across the whole service – my rudimentary understanding is that the new one is a more literal translation and includes the "Get" bit, so is probably more in line with the standard – but as it's quite a big change, I think it is worth double-checking.

jaucourt (approved):

irisfaraway (approved):

PR #1946, opened 2024-04-15
RP job local failure

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3840

Adding delay too recurring payment job so is able to access Sales API when ran


jaucourt (approved):
The delay is perhaps a little crude, ideally we'd test for whether the Sales API is available, but whatever works.

When you squash and merge, please can you make sure the commit message is in the normal format:

```
Summary

Ticket link

Full description
```

As you look to have missed the blank first commit with that format on this PR. Thanks!

PR #1949, opened 2024-04-25
Licence conditions match Notify

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3994

Update licence conditions to match the content on notify letter


packages/gafl-webapp-service/src/pages/terms-and-conditions/__tests__/route.spec.js
```diff
@@ -14,13 +14,50 @@ jest.mock('../../../processors/mapping-constants.js', () => ({
     'salmon-and-sea-trout': 'salmon_and_sea_trout'
   }
 }))
+jest.mock('@defra-fish/business-rules-lib', () => ({
+  FULFILMENT_SWITCHOVER_DATE: '2024-05-30T23:00:00.000Z'
```
jaucourt, line 18: This should be set to a different value than what's in business rules lib, otherwise what's the point of mocking it? When you mock it, it's supposed to show that it's decoupled, which setting to the same value won't do.

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/terms-and-conditions/terms-and-conditions.njk
```diff
@@ -56,7 +72,7 @@
                   items: [
                     {
                       value: "yes",
-                      text: mssgs.terms_conds_item_agree
```
jaucourt, line 59: Now you've removed this, please can you check whether it can be deleted from the language resource files?

packages/gafl-webapp-service/src/pages/terms-and-conditions/__tests__/route.spec.js
```diff
@@ -71,4 +107,9 @@ describe('terms-and-conditions get data', () => {
     const request = getMockRequest({ contactSummary: false })
     expect(() => getData(request)).rejects.toThrowRedirectTo(CONTACT_SUMMARY.uri)
   })
+
+  it.each([[true], [false]])('content returns notify content when isLicenceForYou equals %s', async licenceForYou => {
+    const { content } = await getData(getMockRequest({ isLicenceForYou: licenceForYou }))
```
jaucourt, line 112: Small one again, but if you change `licenceForYou` to `isLicenceForYou`, you can do `{ isLicenceForYou }` rather than `{ isLicenceForYou: licenceForYou }`

jaucourt (changes requested):
One small change then this can merge

jaucourt (approved):

PR #1953, opened 2024-05-13
Update Node version prerequisite in README

lailien3:
Node should be v14+, up to and including v18

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Mono-repo for the rod licensing digital services.
 
 ## Prerequisites
 
-- Node v14.13+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ Mono-repo for the rod licensing digital services.
 
 ## Prerequisites
 
-- Node v14.13+
+- Node v14.13 - v18 (inclusive)
```
jaucourt, line 26: Actually, we shouldn't be using v14, so change this to `v18.x`, please

jaucourt (changes requested):

jaucourt (approved):

PR #1957, opened 2024-05-16
Recurring payments job sets up new permissions in DynamoDb

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-4017

When the RP job is run and relevant permissions are found that need to be renewed, we want to create a new permission for the renewal and store it in DynamoDb. This permission should have all the relevant data that needs to be carried over. The new permission won't be moved over to the CRM until payment has successfully been taken.

As part of this we need to adapt some of the code that was written for easy renewals.


packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -40,11 +47,123 @@ describe('recurring-payments-processor', () => {
...
+          }
+        ],
+        [
+          {
+            dataSource: 'Recurring Payment',
+            permissions: [expect.objectContaining({ permitId: secondPermit })]
+          }
+        ]
+      ])
+    })
```
jaucourt, line 167: I think that, rather than testing with two, you could parameterise the describe so that two, three and ten recurring payments are tested in both test cases. Happy to pair up and work on this together

packages/sales-api-service/src/server/routes/__tests__/renewals.spec.js
```diff
@@ -0,0 +1,111 @@
...
+    permissionForFullReferenceNumber.mockResolvedValueOnce({ filter: 'coffee' })
+    const error = 'Wuh-oh!'
+    executeQuery.mockImplementationOnce(() => {
+      throw new Error(error)
+    })
+    const request = getMockRequest({})
+    const responseToolkit = getMockResponseToolkit()
+
+    await expect(permissionRenewalData[0].options.handler(request, responseToolkit)).rejects.toThrow(error)
+  })
```
jaucourt, line 110: to cover the other parts of the export from renewals, I think you could use a snapshot test covering description, notes, tags, etc

packages/sales-api-service/src/services/renewals/renewals.service.js
```diff
@@ -0,0 +1,101 @@
...
+}
+
+const prepareBasePermissionData = existingPermission => {
+  return {
+    isRenewal: true,
+    licenceLength: '12M', // Always for renewals
+    licenceType: existingPermission.permit.permitSubtype.label,
+    numberOfRods: existingPermission.permit.numberOfRods.toString(),
+    isLicenceForYou: true
+  }
```
jaucourt, line 27: this and the previous function would be slightly more succinct if you omitted the return - so do
```
export const preparePermissionDataForRenewal = existingPermission => ({
    ...prepareBasePermissionData(existingPermission),
    ...prepareDateData(existingPermission),
    licensee: prepareLicenseeData(existingPermission),
    permitId: preparePermitId(existingPermission)
})
```

packages/sales-api-service/src/services/renewals/renewals.service.js
```diff
@@ -0,0 +1,101 @@
...
+    renewedHasExpired: false,
+    licenceToStart: licenceToStart.ANOTHER_DATE,
+    licenceStartDate: licenceStartDate.format(cacheDateFormat),
+    licenceStartTime: licenceStartDate.hours()
+  }
+}
+
+const preparePermitId = existingPermission => {
+  return existingPermission.permit.id
+}
```
jaucourt, line 101: Almost all of the above (barring prepareLicenseeData, prepareDateData and dateDataIfNotExpired) could be rewritten without the return statement

jaucourt (changes requested):

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -6,7 +8,49 @@ export const processRecurringPayments = async () => {
...
+const processPermissionData = async referenceNumber => {
+  console.log('Preparing data based on', referenceNumber)
+  const data = await salesApi.preparePermissionDataForRenewal(referenceNumber)
+  const transactionData = {
+    dataSource: 'Recurring Payment',
+    permissions: [
+      {
+        isLicenceForYou: data.isLicenceForYou,
+        isRenewal: data.isRenewal,
+        issueDate: null,
```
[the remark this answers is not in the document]
irisfaraway: Required by the permission schema, unfortunately!

jaucourt (approved):

PR #1979, opened 2024-06-12
GTM activiated by user

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4081

banner activates GTM if approved


packages/gafl-webapp-service/src/handlers/page-handler.js
```diff
@@ -132,7 +132,6 @@ export default (path, view, completion, getData) => ({
     pageData.analyticsMessageDisplayed = retrieveKeyOrFalse(analytics, ANALYTICS.seenMessage)
     pageData.analyticsSelected = retrieveKeyOrFalse(analytics, ANALYTICS.selected)
     pageData.acceptedTracking = retrieveKeyOrFalse(analytics, ANALYTICS.acceptTracking)
-    pageData.gtmContainerId = process.env.GTM_CONTAINER_ID || false
```
irisfaraway, line 135: We still need this value as it's included in the GTM script on the page (see lines 175 and 177 of the njk file)

packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -1,11 +1,31 @@
...
+  const canTrack = await trackAnalyticsAccepted(request, pageOmit)
+  const optDebug = process.env.ENABLE_ANALYTICS_OPT_IN_DEBUGGING?.toLowerCase() === 'true'
+  const gtmContainerId = process.env.GTM_CONTAINER_ID
+  if (optDebug && gtmContainerId) {
+    if (canTrack === true) {
+      debug(`GTM Container Id: ${gtmContainerId} is being tracked`)
+    } else if (pageOmit === true) {
+      debug(`GTM Container Id: ${gtmContainerId} is not being tracked for current page`)
+    } else {
+      debug(`GTM Container Id: ${gtmContainerId} is not being tracked`)
```
irisfaraway, line 23: The logging here doesn't quite match up with what we had previously – it used to be that we would log if the session was being tracked, and that's what it looks like you are still intending from the tests. However this will just output the same GTM container ID each time, as that is a set thing for the environment.

irisfaraway (changes requested):

irisfaraway (approved):

PR #1983, opened 2024-06-19
Recurring payment job concessions

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3916

automatically transition from full licence to senior concessionary licence when it renews after my 66th birthday


nabeelamir-defra (approved):

packages/sales-api-service/src/services/renewals/renewals.service.js
```diff
@@ -83,4 +90,20 @@ const dateDataIfNotExpired = endDateMoment => {
...
+  delete existingPermission.licensee.noLicenceRequired
+  const ageAtLicenceStartDate = moment(dateData.licenceStartDate)
+    .add(1, 'year')
+    .diff(moment(existingPermission.licensee.birthDate), 'years')
+
+  if (isSenior(ageAtLicenceStartDate)) {
+    await addSenior(existingPermission)
+  }
+
+  return existingPermission
```
irisfaraway, line 108: Is there a reason to not just return `existingPermission.concessions` given that that's the only data we are using from it? Would make `preparePermissionDataForRenewal()` a bit less complex too.

packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
```diff
@@ -28,7 +52,8 @@ describe('preparePermissionDataForRenewal', () => {
       },
       shortTermPreferredMethodOfConfirmation: {
         label: 'Text'
-      }
+      },
+      noLicenceRequired: true
```
irisfaraway, line 56: Feels kinda weird applying this as `true` by default when in practice it should only be true on junior licences. If we have a test for juniors then I'd rather override it then and leave it false the rest of the time.

packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
```diff
@@ -107,7 +136,42 @@ describe('preparePermissionDataForRenewal', () => {
...
+  describe('prepareConcessionsData', () => {
+    it('should add senior concession if the licensee is senior', async () => {
+      const samplePermission = existingPermission()
+      samplePermission.licensee.birthDate = '1954-01-01'
+      const ppd = await preparePermissionDataForRenewal(samplePermission)
+      const senior = { name: 'Senior', id: 'd0ece997-ef65-e611-80dc-c4346bad4004', proof: { type: 'No Proof' } }
+      expect(ppd.concessions[0]).toEqual(senior)
+    })
+
+    it('should remove senior concession if the licensee is not senior', async () => {
```
irisfaraway, line 152: Technically this isn't removing a concession as none exists in the first place – can we rename the test to reflect this?

packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
```diff
@@ -107,7 +136,42 @@ describe('preparePermissionDataForRenewal', () => {
         licenceStartTime: endDate.hours(),
         renewedEndDate: endDate.toISOString()
       }
-      expect(preparePermissionDataForRenewal(existingPermission({ endDate }))).toEqual(expect.objectContaining(expectedData))
+      expect(await preparePermissionDataForRenewal(existingPermission({ endDate }))).toEqual(expect.objectContaining(expectedData))
+    })
+  })
+
+  describe('prepareConcessionsData', () => {
```
irisfaraway, line 143: Could we also add some tests to make sure that disability concessions are also handled appropriately?

packages/sales-api-service/src/services/permit.service.js
```diff
@@ -0,0 +1,41 @@
+import { getReferenceDataForEntity } from './reference-data.service.js'
+import { Permit, Concession, PermitConcession } from '@defra-fish/dynamics-lib'
+
+export const findPermit = async existingPermission => {
+  const licenseeConcessions = existingPermission.concessions || []
+  const permitsJoinPermitConcessions = await getPermitsJoinPermitConcessions()
+
+  if (licenseeConcessions.some(c => !c.name && !!c.type)) {
+    throw new Error("Concession should contain 'name' not 'type'")
```
irisfaraway, line 9: What's the reason for checking for this? Do we have a data quality issue we need to sort out?

packages/sales-api-service/src/services/concession.service.js
```diff
@@ -0,0 +1,61 @@
...
+        type: CONCESSION_PROOF.none
+      }
+    })
+  }
+}
+
+export const removeSenior = permission => removeConcessionType(permission, CONCESSION.SENIOR)
+export const hasSenior = permission => hasConcessionType(permission, CONCESSION.SENIOR)
+
+export const removeJunior = permission => removeConcessionType(permission, CONCESSION.JUNIOR)
```
irisfaraway, line 60: Do we need an `addJunior`? Realise this won't come up for recurring payments but wasn't sure if it'll fall into IWTF-4147

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
```diff
@@ -0,0 +1,283 @@
...
+        const permission = getMockPermission()
+        await concessionService.addDisabled(permission, CONCESSION_PROOF.NI, 'national-insurance-456')
+        expect(permission.concessions).toContainEqual(expect.objectContaining(getDisabledNiConcession()))
+      })
+
+      it('add blue badge to senior licence', async () => {
+        const permission = getMockPermission()
+        permission.concessions = [getSeniorConcession()]
+        await concessionService.addDisabled(permission, CONCESSION_PROOF.blueBadge, 'blue-badge-123')
+        expect(permission.concessions).toContainEqual(getDisabledBlueBadgeConcession())
```
irisfaraway, line 109: Can we also test that the senior concession is still there, like we do when adding a senior concession to a disabled concession (so the other way round)?

Same goes for test on line 116.

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
```diff
@@ -0,0 +1,283 @@
...
+
+      it('add NI to senior licence', async () => {
+        const permission = getMockPermission()
+        permission.concessions = [getSeniorConcession()]
+        await concessionService.addDisabled(permission, CONCESSION_PROOF.NI, 'national-insurance-456')
+        expect(permission.concessions).toContainEqual(getDisabledNiConcession())
+      })
+    })
+
+    describe('removeDisabled', () => {
```
irisfaraway, line 120: Do we need to test that any other concessions are not also removed (so if the permission also has a senior concession and a disabled concession, removing the disabled one does not affect the senior)? Same goes for tests on the other `remove` functions.

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
```diff
@@ -0,0 +1,283 @@
...
+
+    describe('hasSenior', () => {
+      it('is true if licensee has senior concession', async () => {
+        const permission = getMockPermission()
+        permission.concessions = [getSeniorConcession()]
+        const isSenior = await concessionService.hasSenior(permission)
+        expect(isSenior).toBe(true)
+      })
+
+      it("is false if licensee doesn't have senior concession", async () => {
```
irisfaraway, line 229: Maybe overkill, but as we are only testing that it is false if there are no concessions, can we also test that it is false if there is only a non-senior concession? Same goes for tests on the other `has` functions.

packages/recurring-payments-job/package-lock.json
```diff
@@ -46,9 +46,9 @@
       }
     },
     "node_modules/@defra-fish/business-rules-lib": {
-      "version": "1.47.0-rc.7",
-      "resolved": "https://registry.npmjs.org/@defra-fish/business-rules-lib/-/business-rules-lib-1.47.0-rc.7.tgz",
-      "integrity": "sha512-6uQ7vS1nYI9GHVgmGCC82Qgcn1b1MumQRWlEmytKXoj9d+BNcp0wVovaoaAgnev+71O5wvPYe9rYx4ucET5CEA==",
+      "version": "1.47.0-rc.12",
```
irisfaraway, line 49: Looks like the version thing is still an issue :/ Might have to manually bump this up to v1.49.0-rc.3 or whatever the latest version is at time of merging.

irisfaraway (changes requested):
Well, I can safely say that "concession" no longer looks like a word after this 😂 Some thoughts but nothing major

packages/sales-api-service/src/services/permit.service.js
jaucourt: I put it in because I've seen a lot of usage of 'type' rather than 'name', a hold-over (I think) of copy-and-paste from the GAFL easy renewals. The recurring payments code now uses name rather than type, but I thought I'd put in an error as if type is specified in concessions rather than name, the problem is not at all obvious and may well trip us up when applying to Easy Renewals.

packages/sales-api-service/src/services/concession.service.js
jaucourt: We don't for recurring payments, so I've left it to be implemented when / if it's needed.

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
jaucourt: Weirdly, I'd already done this for other concessions, just not for disabled concessions

packages/recurring-payments-job/package-lock.json
jaucourt: I'm not sure what's going on with this - I've now committed the latest version from develop, and it's still showing up as changed. I'll ensure that when I merge it doesn't differ from develop.

irisfaraway (approved):

PR #1989, opened 2024-07-09
Hide RCP option for Junior licence purchases

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4066

Anglers under 17 y/o should not be offered the option to set up recurring payments, so it the option has been hidden.

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
@@ -27,23 +27,66 @@ describe('recurringPayReminderDisplay', () => {
   })
 })
 
+// describe('validForRecurringPayment', () => {
+//   it.each([
+//     [true, '12M', true, true, 'not telesales', 18],
+//     [false, '8D', true, true, 'not telesales', 18],
+//     [false, '12M', false, true, 'not telesales', 18],
+//     [false, '12M', true, false, 'not telesales', 18],
+//     [false, '12M', true, true, 'telesales', 18],
+//     [false, '12M', true, true, 'telesales', 16],
+//   ])(
+//     'should return %s as licence length is %s, licence for you is %s and SHOW_RECURRING_PAYMENTS is %s and journey is %s and age is %s',
+//     (expected, length, licenceFor, recurring, telesales, age) => {
+//       process.env.CHANNEL = telesales
+//       process.env.SHOW_RECURRING_PAYMENTS = recurring
+//       const permission = getPermission({ licenceFor, length, age })
+//       const result = validForRecurringPayment(permission)
+//       expect(result).toEqual(expected)
+//     }
+//   )
+// })
+
 describe('validForRecurringPayment', () => {
-  it.each`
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/processors/recurring-pay-helper.js
```diff
@@ -1,5 +1,7 @@
 import { HOW_CONTACTED } from './mapping-constants.js'
 
+const MIN_AGE_FOR_RECURRING_PAYMENT = 18
```
jaucourt, line 3: As discussed on Slack, this shouldn't be here. Instead, you should use the `isMinor`, `isJunior` and `isSenior` functions in business-rules-lib

jaucourt (changes requested):
A few modifications are required, please

packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
```diff
@@ -27,18 +28,20 @@ describe('recurringPayReminderDisplay', () => {
...
+  test.each`
+    expected  | length   | licenceFor | recurring | telesales           | age
+    ${true}   | ${'12M'} | ${true}    | ${true}   | ${'not telesales'}  | ${18}
+    ${false}  | ${'8D'}  | ${true}    | ${true}   | ${'not telesales'}  | ${18}
+    ${false}  | ${'12M'} | ${false}   | ${true}   | ${'not telesales'}  | ${18}
+    ${false}  | ${'12M'} | ${true}    | ${false}  | ${'not telesales'}  | ${18}
+    ${false}  | ${'12M'} | ${true}    | ${true}   | ${'telesales'}      | ${18}
+    ${false}  | ${'12M'} | ${true}    | ${true}   | ${'not telesales'}  | ${17}
+  `(
+    'should return %s as licence length is %s, licence for you is %s and SHOW_RECURRING_PAYMENTS is %s and journey is %s and age is %s',
```
jaucourt, line 40: Too many 'ands' in one sentence ;)

packages/gafl-webapp-service/src/processors/recurring-pay-helper.js
```diff
@@ -13,4 +14,5 @@ export const validForRecurringPayment = permission =>
   process.env.SHOW_RECURRING_PAYMENTS?.toLowerCase() === 'true' &&
   permission.licenceLength === '12M' &&
   permission.isLicenceForYou &&
+  permission.licensee.age > JUNIOR_MAX_AGE + 1 &&
```
jaucourt, line 17: The main problem is that there is no `age` field for licensee, only a `birthDate` field so age will have to be calculated from that.

Furthermore, if age is greater than JUNIOR_MAX_AGE + 1, that'll be 17, meaning recurring payments won't be offered until 18 (as your tests show) when the angler paid for an adult licence the previous year. If you get rid of the `+1` I think this would work, so `age > JUNIOR_MAX_AGE`

jaucourt (changes requested):

jaucourt (approved):
This is great, thanks for persevering!

PR #1995, opened 2024-07-11
Copy and paste into input fields

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4146

prevent users from copying and pasting non standard fonts into input fields

### DIFF
--- a/packages/business-rules-lib/src/validators/contact.validators.js
+++ b/packages/business-rules-lib/src/validators/contact.validators.js
@@ -80,11 +80,8 @@ const createDateStringValidator = joi =>
 
 const allowedUnicodeBlocks = '\\u0000-\\u024F'
 const forbiddenCharsRegex = new RegExp(`[^${allowedUnicodeBlocks}\\s'’()-]`, 'gu')
-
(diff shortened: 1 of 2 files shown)


packages/business-rules-lib/src/validators/contact.validators.js
```diff
@@ -154,15 +191,33 @@ export const overseasPostcodeRegex = /^([a-zA-Z0-9 ]{1,12})$/i
...
 
 /**
  * Create a validator to check/format overseas postcodes
  * @param {Joi.Root} joi the joi validator used by the consuming project
  * @returns {Joi.StringSchema}
  */
 export const createOverseasPostcodeValidator = joi =>
-  joi.string().trim().min(1).max(12).uppercase().required().pattern(overseasPostcodeRegex)
+  checkCopyPasteValidator(joi, forbiddenCharsNumbersRegex)
+    .string()
```
irisfaraway, line 213: Do we need to include `string()` on these if `checkCopyPasteValidator` is already extending the default string validator? Same question goes for the other uses of it.

irisfaraway (commented):
Looks good! Just one more question from me...

irisfaraway (approved):

PR #2010, opened 2024-08-11
Add new validators for date fields

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3913

This adds additional validators for the individual date fields and the overall date to make sure that any dates entered are valid.

An additional PR will follow to apply these validations to the GAFL frontend.

### DIFF
--- a/packages/business-rules-lib/src/validators/__tests__/date.validators.spec.js
+++ b/packages/business-rules-lib/src/validators/__tests__/date.validators.spec.js
@@ -1,0 +1,95 @@
+import Joi from 'joi'
+import * as dateValidation from '../date.validators.js'
+
+describe('date validators', () => {
+  describe('dayValidator', () => {
+    it.each([1, 15, 31])('expects an integer between 1 and 31', async value => {
+      await expect(dateValidation.createDayValidator(Joi).validateAsync(value)).resolves.toEqual(value)
+    })
+
+    it.each([
+      [0, 'must be greater than or equal to 1'],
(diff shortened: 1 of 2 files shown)


packages/business-rules-lib/src/validators/__tests__/date.validators.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+    it.each([
+      [0, 'must be greater than or equal to 1'],
+      [32, 'must be less than or equal to 31'],
+      [1.5, 'must be an integer'],
+      ['foo', 'must be a number'],
+      ['', 'must be a number'],
+      [null, 'must be a number'],
+      [undefined, 'is required']
+    ])('throws an error if the value is invalid', async (value, message) => {
```
jaucourt, line 18: With a test description like this, we'll only have an output of 'throws an error if the value is invalid' seven times, which isn't very informative. A description of 'When date value is %s validator throws an error of %s' would probably do the trick.

This goes for all the other it.each tests in this file, if you don't mind? Thanks

jaucourt (changes requested):
Some changes needed so that we have unique test names, please

packages/business-rules-lib/src/validators/__tests__/date.validators.spec.js
irisfaraway: Sorted!

jaucourt (approved):

PR #2022, opened 2024-08-19
Apply more specific date error messages

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-3913

This PR makes use of the new date validations to give more specific date error messages.


packages/gafl-webapp-service/src/pages/licence-details/licence-to-start/__tests__/route.spec.js
```diff
@@ -27,5 +43,34 @@ describe('licence-to-start > route', () => {
       const result = await getData(request)
       expect(result.isLicenceForYou).toBeFalsy()
     })
+
+    it.each([
```
irisfaraway, line 47: Can we have a test for when there are multiple errors?

irisfaraway (commented):
I can't officially approve or reject this since it's technically my PR (although looks like it's all your work!) so someone else will have to stamp this as well.

Do we need to clean up the new date validators in the business-lib since we haven't ended up using them? (createDayValidator, etc)

packages/gafl-webapp-service/src/schema/validators/validators.js
```diff
@@ -26,18 +33,33 @@ export const startDateValidator = payload => {
...
-      .startOf('day')
-      .toDate()
-    Joi.assert({ startDate }, Joi.object({ startDate: Joi.date().min(minDate.toDate()).max(maxDate.toDate()) }))
+    const minDate = moment().tz(SERVICE_LOCAL_TIME).startOf('day').toDate()
+    const maxDate = moment().tz(SERVICE_LOCAL_TIME).add(ADVANCED_PURCHASE_MAX_DAYS, 'days').toDate()
+    validateDate(day, month, year, minDate, maxDate)
   }
 }
+
+export const getErrorFlags = error => {
```
irisfaraway, line 46: Suggest naming this `getDateErrorFlags` just so the specific purpose is more clear.

irisfaraway (commented):
Much tidier Nunjucks files, yay! I have one nitpick about naming and then I'm happy.

PR #2025, opened 2024-08-22
Cookies page analytics

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4236

functionality for accepting/reject analytics from cookies page


packages/gafl-webapp-service/src/handlers/analytics-handler.js
```diff
@@ -49,23 +49,6 @@ export const pageOmitted = async request => {
 }
 
 export default async (request, h) => {
-  await checkAnalyticsResponse(request)
```
irisfaraway, line 52: Maybe I'm missing something but what is the reason for removing `checkAnalyticsResponse()` and moving that logic into the default function? The function is getting quite long as a result.

packages/gafl-webapp-service/src/handlers/__tests__/analytics-handler.spec.js
```diff
@@ -163,6 +163,36 @@ describe('The analytics handler', () => {
     })
   })
 
+  describe('checkAnalyticsCookiesPage', () => {
+    it.each([
+      [true, false, true, 'reject'],
+      [true, true, true, 'accept']
+    ])(
+      'sets selected to %s, acceptTracking to %s and seenMessage to %s when analyticsResponse is %s',
```
irisfaraway, line 171: Might be good to have a test to make sure nothing is updated if a value that isn't `accept` or `reject` is somehow submitted.

irisfaraway (changes requested):
Just some minor questions!

packages/gafl-webapp-service/src/handlers/analytics-handler.js
ScottDormand96: I moved back to what was previous and will use this for when using banner at top of page, created new function underneath for specific too cookies page as needs to do it slightly different

irisfaraway (approved):

PR #2028, opened 2024-08-28
Prevent copy and paste values

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4146

Stop users from copy and pastign different values into input fields


packages/gafl-webapp-service/src/locales/cy.json
```diff
@@ -781,7 +784,7 @@
   "terms_conds_bulletpoint_6_notify_bobo": "cyflwyno datganiad blynyddol, hyd yn oed os nad yw wedi dal unrhyw eogiaid na brithyllod y môr, neu os na fu'n pysgota - gellir ",
   "terms_conds_bulletpoint_6_link_notify_bobo": "cyflwyno datganiad ar-lein (yn agor mewn tab newydd)",
   "terms_conds_title_notify_bobo": "A yw deiliad y drwydded yn cytuno ag amodau’r drwydded?",
-  "there_is_a_problem": "Mae problem",
+  "there_is_a_problem": "Mae problem wedi codi",
```
irisfaraway, line 787: This will change the error text throughout the app – is that something we want to be doing as part of this ticket?

irisfaraway (commented):
Just have a question about a Welsh language change.

packages/gafl-webapp-service/src/locales/cy.json
lailien3: Probably not, I can either leave it as "Mae problem" or create a new variable, wdyt?

irisfaraway: I would say just leave it as-is but maybe check with Jen when she's back?

PR #2035, opened 2024-09-04
Recurring payments dependencies not updating

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4117

Package-lock.json node modules are not updating for recurring payments


jaucourt (commented):
This looks good but I'm interested in why it's always worked up to now without the specific lerna exec to update package.lock.json files. You may well not know this!

PR #2045, opened 2024-09-12
Display Recurring payment terms and conditions

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-3305

View recurring payments via hyperlink without having to go through the journey

### DIFF
--- a/packages/gafl-webapp-service/src/__tests__/server.spec.js
+++ b/packages/gafl-webapp-service/src/__tests__/server.spec.js
@@ -15,7 +15,8 @@ jest.mock('../uri.js', () => ({
   PRIVACY_POLICY: { uri: '/PRIVACY_POLICY' },
   REFUND_POLICY: { uri: '/REFUND_POLICY' },
   NEW_TRANSACTION: { uri: '/NEW_TRANSACTION' },
-  NEW_PRICES: { uri: '/NEW_PRICES' }
+  NEW_PRICES: { uri: '/NEW_PRICES' },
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/__tests__/server.spec.js
```diff
@@ -52,31 +53,37 @@ describe('The server', () => {
     expect(serverDecorateSpy).toHaveBeenCalledWith('toolkit', 'redirectWithLanguageCode', expect.any(Function))
   })
 
-  it('addLanguageCodeToUri is called with request but not new prices', async () => {
-    createServer(catboxOptions)
-    const serverDecorateSpy = jest.spyOn(server, 'decorate').mockImplementation(() => {})
+  it.each([[uris.NEW_PRICES.uri], [uris.RECURRING_TERMS_CONDITIONS.uri]])(
+    'addLanguageCodeToUri is called with request but not %s',
```
irisfaraway, line 57: I realise this is adapting an existing test but I am not sure I understand the point of it – if we are hitting `/buy` I'm not sure why it would ever be calling `addLanguageCodeToUri` with these URIs? It might make more sense to hit these URIs and then make sure we are not calling `addLanguageCodeToUri` with a redirect.

irisfaraway (changes requested):

irisfaraway (approved):

jaucourt (approved):

PR #2049, opened 2024-09-17
Add query to allow angler to login in to RCR

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-2715

Add a query to dynamics-lib which allows an angler to login using their permission reference number and postcode


packages/dynamics-lib/src/queries/__tests__/contact.queries.spec.js
```diff
@@ -0,0 +1,98 @@
...
+      expect(result.DateOfBirth).toBeNull()
+      expect(result.Premises).toBeNull()
+      expect(result.Street).toBeNull()
+      expect(result.Town).toBeNull()
+      expect(result.Locality).toBeNull()
+      expect(result.Postcode).toBeNull()
+      expect(result.ReturnStatus).toBe('error')
+      expect(result.SuccessMessage).toBe('')
+      expect(result.ErrorMessage).toBe('contact does not exists')
+      expect(result.ReturnPermissionNumber).toBeNull()
```
jaucourt, line 95: One test, thirteen assertions? Tut, tut 😛
I appreciate that the setup is identical, however please can we refactor this to use it.each and a good description - or perhaps a snapshot. Either way, avoid more than one assertion in a test please.

jaucourt (changes requested):

packages/dynamics-lib/src/queries/__tests__/contact.queries.spec.js
nabeelamir-defra: I've changed this to use `toMatchObject`!

jaucourt (approved):

PR #2060, opened 2024-10-01
Update dependencies in package-lock

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4117

update deploy script to update package-lock files

### DIFF
--- a/scripts/github/deploy.sh
+++ b/scripts/github/deploy.sh
@@ -67,6 +67,11 @@ echo "Updating version from ${PREVIOUS_VERSION} to ${NEW_VERSION}"
 # Update package files versions, project inter-dependencies and lerna.json with new version number
 lerna version "${NEW_VERSION}" --yes --no-push --force-publish --exact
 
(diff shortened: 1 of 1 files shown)


scripts/github/deploy.sh
```diff
@@ -67,6 +67,11 @@ echo "Updating version from ${PREVIOUS_VERSION} to ${NEW_VERSION}"
 # Update package files versions, project inter-dependencies and lerna.json with new version number
 lerna version "${NEW_VERSION}" --yes --no-push --force-publish --exact
 
+echo "Updating package-lock.json from ${PREVIOUS_VERSION} to ${NEW_VERSION}"
+# Update package-lock.json files to reflect new versions
+lerna exec -- npm install --package-lock-only
```
jaucourt, line 72: How's this different to what you did before (not being snarky, genuine question)?

ScottDormand96: change in order and updated echo so same lets know version updated with. Now once package.json updated it will update package-lock straight after with same version then deploy

jaucourt (approved):

PR #2062, opened 2024-10-06
Upgrade AWS SDK DynamoDB from v2 > v3

lailien3:
https://eaflood.atlassian.net/browse/IWTF-3449

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,11 @@
       "version": "0.0.1",
       "hasInstallScript": true,
       "license": "SEE LICENSE IN LICENSE",
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


jaucourt (changes requested):
I've reviewed some of this, and I'm noticing common issues, such as use of `let` rather than `const`, multiple `expect` statements in tests and enforcement of a particular implementation via tests, rather than testing an expected result after particular inputs. Please could you check through yourself for further instances of these (I've highlighted the first few). Also, you're touching areas of the codebase that haven't been looked at since it was first written, and I'd say this is an opportunity to rewrite the over-complicated tests into the style we use now. Have a look through yourself and then we can go through it together?

PR #2064, opened 2024-10-09
Remove ftp functionality

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4278

Remove ftp functionality as package we use in fulfilment and pocl jobs (ssh2-sftp-client) has critical vulnerability


packages/pocl-job/src/io/s3.js
```diff
@@ -47,13 +44,6 @@ export const refreshS3Metadata = async () => {
     const filename = file.Key.split('/').pop()
 
     console.log(`Processing ${filename}`)
-
-    const dynamicsRecord = await salesApi.getTransactionFile(filename)
-    if (!dynamicsRecord || !DYNAMICS_IMPORT_STAGE.isAlreadyProcessed(dynamicsRecord.status.description)) {
-      await storeS3Metadata(file.ETag, filesize(file.Size), filename, file.Key, moment(new Date(file.LastModified)))
-    } else {
-      console.log(`${file.Key} is already processed, skipping`)
-    }
```
irisfaraway, line 56: Not sure if we want to delete this and the `storeS3Metadata` function – I think this can still be used for other file uploads as well and we don't want to blow up the direct debit process. However if we do keep it then it might make sense to move that function out of a file called ftp-to-s3.js anyway.

irisfaraway (changes requested):
Now that's the kind of line count diff I like to see, hahaha.

One suggested change to do with the DD process.

jaucourt (approved):
LGTM, but let's wait until Iris is back and can re-review than the s3 changes have been made to her satisfaction

irisfaraway (approved):

irisfaraway (approved):

packages/pocl-job/src/config.js
```diff
@@ -1,46 +1,3 @@
...
-  'aes192-cbc',
-  'aes128-cbc',
-  'blowfish-cbc',
-  '3des-cbc',
-  'cast128-cbc'
-]
-
 class Config {
   _db
   _ftp
```
jaucourt, line 3: I don't think this is needed anymore either, is it?

jaucourt (changes requested):
One little thing to change

PR #2069, opened 2024-10-24
Change docker to not install dev dependency

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4280

Change dockerfile to run npm ci instead of npm i so dev dependency not included in docker container


irisfaraway (approved):
Interesting that we were already doing this on some packages!

PR #2074, opened 2024-11-11
Add agreement_id to existing GOV.UK Pay request

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-4261

If the request to prepare a payment is part of setting up a new recurring payment agreement, then the agreement_id should be included.

I also renamed prepareRecurringPayment() to prepareRecurringPaymentAgreement() as I think this is slightly more accurate - it is setting up the agreement rather than the payment - and it initially threw me off when working out what needed to be changed.


packages/gafl-webapp-service/src/handlers/__tests__/agreed-handler-recurring-payments.spec.js
```diff
@@ -32,12 +38,13 @@ describe('The agreed handler', () => {
...
   beforeEach(jest.clearAllMocks)
 
+  const mockTransactionCacheSet = jest.fn()
   const getMockRequest = (overrides = {}) => ({
     cache: () => ({
       helpers: {
         transaction: {
           get: async () => ({ cost: 0 }),
-          set: async () => {}
+          set: mockTransactionCacheSet
```
jaucourt, line 47: This shouldn't be referring to an externally defined variable, please  define a new parameter. For the avoidance of confusion, it might be best to have an object config for the mock, e.g.
```
const getMockRequest = ({
  overrides = {},
  mockTransactionCacheSet = () => {}
} = {}) => ({
```
(I put mockTransactionCacheSet as a plain arrow function as the functionality of jest.fn isn't required most of the time)

packages/gafl-webapp-service/src/processors/payment.js
```diff
@@ -50,17 +50,21 @@ export const preparePayment = (request, transaction) => {
     }
   }
 
+  if (recurring) {
+    result.set_up_agreement = transaction.agreementId
+  }
```
jaucourt, line 55: Could we modify the logic a bit - so that if agreementId is present, it is included, rather than relying on the flag variable? Seems like it might be a bit of a 'gotcha' further down the line.

packages/sales-api-service/src/schema/transaction.schema.js
```diff
@@ -33,7 +35,8 @@ const createTransactionRequestSchemaContent = {
   }),
   createdBy: Joi.string().optional(),
   journalId: Joi.string().optional(),
-  transactionId: Joi.string().guid({ version: 'uuidv4' })
+  transactionId: Joi.string().guid({ version: 'uuidv4' }),
+  agreementId: Joi.string().alphanum().length(AGREEMENT_ID_LENGTH).optional()
```
jaucourt, line 39: Would you mind modifying transactionId with the `.optional()` method? All keys are optional by default, which is why I omitted it, but it's included everywhere else and it does make it more readable, so it probably should be here.

jaucourt (changes requested):

jaucourt (approved):

PR #2090, opened 2024-12-10
Mobile phone links

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4374

Mobile phone numbers are appearing in analytics since implementing tracking change links, remove as clickable link

### DIFF
--- a/packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
+++ b/packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
@@ -70,7 +70,7 @@ class RowGenerator {
 
   _generateRow (label, text, rawHref, visuallyHiddenText, id) {
     const href = addLanguageCodeToUri(this.request, rawHref)
-    return {
+    const row = {
       key: {
         text: label
       },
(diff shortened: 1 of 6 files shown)


packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
```diff
@@ -90,6 +90,11 @@ class RowGenerator {
         }
       })
     }
+    if (text.startsWith('Text message')) {
+      row.value.meta = '<meta name="format-detection" content="telephone=no">'
+    }
```
jaucourt, line 95: This won't work if the language is set to Welsh. It's also pretty brittle to determine whether to include the meta tag based on a specific text string being present.

 Rather than doing this, an additional boolean parameter flagging whether or not to add the meta value should be added. I think this pushes the number of arguments over the line into making them an object rather than an additional parameter, but make the flag for including the meta tag optional, defaulting to false, so that only the `generateContactRow` function needs to include it. Furthermore, the html should be in a const variable defined at the top of the file.

jaucourt (changes requested):

jaucourt (approved):

PR #2091, opened 2024-12-11
Update activity in CRM

nabeelamir-defra:
https://eaflood.atlassian.net/browse/IWTF-4401

Update activity in CRM

### DIFF
--- a/packages/dynamics-lib/src/queries/__tests__/activity.queries.spec.js
+++ b/packages/dynamics-lib/src/queries/__tests__/activity.queries.spec.js
@@ -1,4 +1,4 @@
-import { createActivity } from '../activity.queries.js'
+import { createActivity, updateActivity } from '../activity.queries.js'
 import { dynamicsClient } from '../../client/dynamics-client.js'
 
 jest.mock('dynamics-web-api', () => {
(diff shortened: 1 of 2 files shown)


packages/dynamics-lib/src/queries/__tests__/activity.queries.spec.js
```diff
@@ -69,4 +69,63 @@ describe('Activity Service', () => {
...
+    }
+
+    const errorResponse = {
+      '@odata.context': 'https://dynamics.om/api/data/v9.1/defra_UpdateRCRActivityResponse',
+      RCRActivityId: null,
+      ReturnStatus: 'error',
+      SuccessMessage: '',
+      ErrorMessage: 'Failed to update activity',
+      oDataContext: 'https://dynamics.com/api/data/v9.1/defra_UpdateRCRActivityResponse'
+    }
```
jaucourt, line 89: Please can you convert both of these from variables to functions, so that a fresh copy of the data is obtained each time? Also, change the name to `getMockResponse` / `getErrorResponse`

jaucourt (changes requested):

PR #2101, opened 2024-12-23
Move continue button to below the pricing summary

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4379
https://eaflood.atlassian.net/browse/IWTF-4380

Pricing summary appears beneath green button.

This means that if a user doesn't scroll down, they will miss them and may not know the licence price to make an informed decisions until further into the journey. This may result in a user clicking back and forth through the journey rather than proceeding smoothly through.

### DIFF
--- a/packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
+++ b/packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
@@ -114,31 +114,17 @@ div.initially-hidden {
   white-space: nowrap;
 }
 
-.page-layout {
-  display: flex;
-  flex-direction: row;
-}
-
-.page-layout__summary {
-  order: 2;
-  flex: 0 0 30%;
-  margin-right: 2em;
-}
-
-.page-layout__main {
-  order: 2;
-  flex: 1;
-}
-
 @media (max-width: 640px) {
-  .page-layout {
-    flex-direction: column;
+  .govuk-button {
+    position: absolute;
+    bottom: 0px;
+    left: 15px;
+    right: 15px;
+    width: auto;
   }
 
-  .page-layout__summary,
-  .page-layout__main {
-    flex: none;
-    width: 100%;
-    margin-right: 0;
+  .govuk-grid-row {
+    position: relative;
+    padding-bottom: 75px;
   }
 }
--- a/packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
+++ b/packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
@@ -118,13 +118,13 @@ div.initially-hidden {
   #continue {
     position: absolute;
     top: auto;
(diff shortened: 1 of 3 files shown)


packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
```diff
@@ -113,3 +113,19 @@ div.initially-hidden {
 .phone-number {
   white-space: nowrap;
 }
+
+@media (max-width: 640px) {
+  #continue {
+    position: absolute;
+    top: auto;
+    bottom: 0px;
```
jaucourt, line 121: no need for `px` here

packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
```diff
@@ -113,3 +113,19 @@ div.initially-hidden {
...
+    left: 15px;
+    right: 15px;
+    width: auto;
+  }
+
+  .govuk-grid-row {
+    position: relative;
+    padding-bottom: 75px;
+  }
```
jaucourt, line 130: These styles are fine, but they'll require far more testing effort than if you confined them to only the pages they're needed on - i.e. the licence type and licence length pages. To do this we can use the :has pseudo-class:
```
.govuk-grid-row:has(.payment-summary-wrap)
```
This will only target the `#continue` and `.govuk-grid-row` elements on pages where the payment summary exists, which is exactly what we want. The :has pseudo class is supported [by mobile browsers released in 2022 and 2023](https://developer.mozilla.org/en-US/docs/Web/CSS/:has), so well within [Grade B support](https://github.com/alphagov/govuk-frontend?tab=readme-ov-file#browser-and-assistive-technology-support). Realistically, there will be those with old smartphones stuck on old versions of Android or iOS that don't offer support to this selector, but they'll regress gracefully to the version where the prices display beneath the continue button. We should check this with the PO after QA to see whether it's acceptable to do this.

jaucourt (changes requested):

packages/gafl-webapp-service/assets/sass/rod-licensing/_general.scss
```diff
@@ -113,3 +113,19 @@ div.initially-hidden {
...
+    bottom: 0;
+    left: 15px;
+    right: 15px;
+    width: auto;
+  }
+
+  .govuk-grid-row:has(.payment-summary-wrap) {
+    position: relative;
+    padding-bottom: 75px;
+  }
```
jaucourt, line 130: I think you've misunderstood my last comment - the #continue button styles need to go inside a parent selector too, otherwise every single continue button will have the same style and so be absolutely positioned, which I don't think is correct.

jaucourt (changes requested):

jaucourt (approved):

PR #2106, opened 2024-12-24
Capitalising Date Entry Fields

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4393

Date entry fields are currently all in lower case across the service. They should begin capitalised.


jaucourt (changes requested):
Totally not your fault, but I think the labels here could be better. I think they should be `date_day`, `date_month` and `date_year`, as they're not only used for date of birth but licence start date too. Also, is the Welsh capitalised? If not, should we do it in this ticket?

lailien3:
> Totally not your fault, but I think the labels here could be better. I think they should be `date_day`, `date_month` and `date_year`, as they're not only used for date of birth but licence start date too. Also, is the Welsh capitalised? If not, should we do it in this ticket?

good shout 👍🏼  Welsh is already capitalised 🕺🏼

PR #2111, opened 2025-01-19
Check recurring status based on transaction

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-4265

When checking on the order complete page if the new licence has a recurring payment agreement, base this off the existence of an agreementId on the transaction rather than trying to get it from the permission.

### DIFF
--- a/packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
+++ b/packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
@@ -58,12 +58,7 @@ const postalFulfilment = permission => {
   }
 }
 
-const isRecurringPayment = transaction => {
-  if (process.env.SHOW_RECURRING_PAYMENTS?.toLowerCase() === 'true' && transaction.agreementId) {
-    return true
-  }
-  return false
-}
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/pages/order-complete/order-complete/route.js
```diff
@@ -58,7 +58,12 @@ const postalFulfilment = permission => {
   }
 }
 
-const isRecurringPayment = permission => process.env.SHOW_RECURRING_PAYMENTS?.toLowerCase() === 'true' && permission.isRecurringPayment
+const isRecurringPayment = transaction => {
+  if (process.env.SHOW_RECURRING_PAYMENTS?.toLowerCase() === 'true' && transaction.agreementId) {
+    return true
+  }
+  return false
+}
```
jaucourt: I think this would be simpler if it were:
```
const isRecurringPayment = transaction => (process.env.SHOW_RECURRING_PAYMENTS?.toLowerCase() === 'true' && !!transaction.agreementId)
```

irisfaraway: Good shout, fixed!

PR #2113, opened 2025-01-22
Validate RP Request

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-4057

Validate the response for recurring payments. This is in packages/sales-api-service/src/server/routes/recurring-payments.js. Add a schema for validating input in similar way to renewals authentication.

### DIFF
--- a/packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
+++ b/packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
@@ -57,20 +57,43 @@ describe('getDueRecurringPaymentsSchema', () => {
     await expect(() => dueRecurringPaymentsResponseSchema.validateAsync(getSampleData())).not.toThrow()
   })
 
-  it.each(['name', 'nextDueDate', 'endDate', 'agreementId'])('throws an error if %s is missing', async property => {
+  it.each([
+    'id',
+    'name',
+    'status',
+    'nextDueDate',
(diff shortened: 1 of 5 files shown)


packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    ['name', 99],
+    ['nextDueDate', 'not-a-date'],
+    ['agreementId', 'not-a-guid'],
+    ['contactId', 'still-not-a-guid']
+  ])('throws an error if %s is not the correct type', async (property, value) => {
+    const sampleData = getSampleData()
+    sampleData[property] = value
+    expect(() => dueRecurringPaymentsResponseSchema.validateAsync(sampleData)).rejects.toThrow()
+  })
+})
```
irisfaraway, line 76: There are some values we aren't testing individually - status, cancelledDate, cancelledReason, activePermission and publicId – and we don't look to be testing that the expanded object has the correct schemas applied either. Is there a reason for not testing these?

jaucourt: I took a sample of several fields rather than testing every single field, similar to other schemas we've done, but you make a fair point about the expanded objects so I'll put in some tests for those

irisfaraway (approved):

PR #2122, opened 2025-02-11
RP job handles GOV.UK Pay error responses

lailien3:

If our request to the payment API returns an error response code (for example, if the API is down or if our authorisation is invalid) then the recurring payment job needs to handle this gracefully. This ticket implements and error catching and logging.

If we receive an error response, this response is outputted to our logs; we abort the run if systems are down at the start of a run and try again; we continue with run if systems are down mid-run and try again; If, after both runs, a payment has still not been taken,  the run on the following day will detect that the window for processing these payments has passed and automatically cancel them.

### DIFF
--- a/packages/recurring-payments-job/src/recurring-payments-processor.js
+++ b/packages/recurring-payments-job/src/recurring-payments-processor.js
@@ -6,7 +6,7 @@ import { getPaymentStatus, sendPayment } from './services/govuk-pay-service.js'
 const PAYMENT_STATUS_DELAY = 60000
 const payments = []
 
-const retry = async (fn, retries = 3, delay = 1000) => {
+export const retry = async (fn, retries = 3, delay = 5000) => {
   try {
     return await fn()
   } catch (error) {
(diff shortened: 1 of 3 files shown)


packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -6,16 +6,33 @@ import { getPaymentStatus, sendPayment } from './services/govuk-pay-service.js'
...
+    return await fn()
+  } catch (error) {
+    if (retries === 0) {
+      throw error
+    }
+    console.log(`Error: ${error}. Retrying... Attempts left: ${retries}`)
+    await new Promise(resolve => setTimeout(resolve, delay))
+    return retry(fn, retries - 1, delay * 2)
+  }
+}
```
jaucourt, line 20: I'm not sure why this is here? The ticket you're working on does mention retrying, but states specifically '“Trying again” means we will have a second run during the day'

jaucourt (changes requested):

packages/recurring-payments-job/src/recurring-payments-processor.js
lailien3: makes sense ty, i overthought the wrong part of the sentence 😅

PR #2133, opened 2025-03-11
Implement batcher for RCP

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-4598

Take the HTTP Batcher in connectors-lib and use it for sending recurring payment requests


packages/connectors-lib/src/govuk-pay-api.js
```diff
@@ -32,6 +32,23 @@ export const createRecurringPaymentAgreement = async preparedPayment => {
   }
 }
 
+export const queueRecurringPayment = (preparedPayment, batcher) => {
+  batcher.addRequest(process.env.GOV_PAY_API_URL, {
```
irisfaraway: Should this be `GOV_PAY_RCP_API_URL`? Same question for `queueRecurringPaymentStatusCheck()` below

packages/connectors-lib/src/__tests__/govuk-pay-api.spec.js
```diff
@@ -140,14 +148,87 @@ describe('govuk-pay-api-connector', () => {
...
-        headers: recurringHeaders,
+        headers: recurringHeaders(),
         method: 'post',
         timeout: 10000
       })
       expect(consoleErrorSpy).toHaveBeenCalled()
     })
   })
 
+  describe('queueRecurringPayment', () => {
```
irisfaraway, line 159: Do we need to test any sort of error handling here, for example if invalid values are passed in or if the batcher itself throws an error?

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
```diff
@@ -175,11 +175,32 @@ describe('HTTP Request Batcher', () => {
...
+        fetch.mockImplementationOnce(() => ({ status: 429 }))
+      }
+      batcher.addRequest('https://api.example.com')
+      await batcher.fetch()
+      expect(fetch).toHaveBeenCalledTimes(maxRequestAttempts)
+    })
+
+    it('stops retrying requests that received a 429 response once a successful response is received', async () => {
+      const batcher = new HTTPRequestBatcher({ batchSize: 1 })
+      batcher.maxRequestAttempts = 5
```
irisfaraway, line 254: Is it worth testing with a higher `maxRequestAttempts` than the number of attempts we expect the test to make? To me, it's not immediately clear if the requests stop because we have stopped mocking the 429 response, or because we have hit the cap on requests.

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
```diff
@@ -232,18 +249,35 @@ describe('HTTP Request Batcher', () => {
...
+      const batcher = new HTTPRequestBatcher()
+      batcher.addRequest('https://api.example.com', { method: 'GET' })
+      batcher.addRequest('https://api.example.com', { method: 'GET' })
+      batcher.addRequest('https://api.example.com', { method: 'GET' })
+      batcher.addRequest('https://api.example.com', { method: 'GET' })
+      batcher.addRequest('https://api.example.com', { method: 'GET' })
+
+      await batcher.fetch()
+
+      expect(fetch).toHaveBeenCalledTimes(5)
```
irisfaraway, line 339: Think it'd be preferable to test which requests have been sent rather than just the number. So maybe adding 5 different requests to the batcher and then checking that each one has been called. That way we know it's 5 different ones being called and not the same one being called again due to a non-200 response.

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -159,47 +202,61 @@ const preparePayment = (agreementId, transaction) => {
...
+    batchSize: Number(process.env.GOV_PAY_GET_BATCH_SIZE),
+    delay: Number(process.env.GOV_PAY_BATCH_DELAY_MS)
+  })
+  for (const payment of payments) {
+    queueRecurringPaymentStatusCheck(payment.paymentId, batcher)
+  }
+  await batcher.fetch()
+
+  return batcher
+}
```
irisfaraway, line 259: Could we add some line breaks between the declarations for readability? Same goes for line 248/9.

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -159,47 +202,61 @@ const preparePayment = (agreementId, transaction) => {
...
-      try {
-        await salesApi.processRPResult(payment.transaction.id, payment.paymentId, payment.created_date)
-        debug(`Processed Recurring Payment for ${payment.transaction.id}`)
-      } catch (err) {
-        console.error(`Failed to process Recurring Payment for ${payment.transaction.id}`, err)
-        throw err
+const checkPaymentStatuses = async payments => {
+  const batcher = await queuePaymentStatusChecks(payments)
+  for (let index = 0; index < batcher.responses.length; index++) {
+    const response = batcher.responses[index]
```
irisfaraway: Could we do this as a `for ... of` loop instead? Bit nicer to read

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+        paymentTimestamp: response.jsonValue.created_date,
+        paymentStatus: PAYMENT_JOURNAL_STATUS_CODES.InProgress
+      })
+    } else {
+      const description = response.jsonValue.description || ''
+      if (
+        description.match(new RegExp(`Invalid attribute value: ${paymentToRequest.agreementId}. Agreement (does not exist|must be active)`))
+      ) {
+        console.log(
+          '%s is an invalid agreementId. Recurring payment %s will be cancelled',
```
irisfaraway, line 158: I think it'd be more helpful to print out a separate `%s is an inactive agreementId` message in that scenario (and then do the cancelling anyway)

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
 
-const takePaymentIfValid = async (preparedPayment, agreementId, transaction) => {
-  try {
-    return await sendPayment(preparedPayment)
-  } catch (error) {
-    if (error.message.includes('Invalid attribute value: agreement_id. Agreement does not exist')) {
-      console.log('%s is an invalid agreementId. Recurring payment %s will be cancelled', agreementId, transaction.recurringPayment.id)
-      await salesApi.cancelRecurringPayment(transaction.recurringPayment.id)
+const processPaymentResponses = async (paymentsToRequest, batcher) => {
+  for (let x = 0; x < paymentsToRequest.length; x++) {
```
irisfaraway: Can we do a `for ... of` loop instead, or do we need to do it this way to get the index for the batcher?

Honestly just matching based on assuming these arrays are in the same order makes me a little nervous. Is there any other common value we can match on?

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+        expanded: {
+          activePermission: {
+            entity: { referenceNumber }
+          }
+        }
+      } = duePayment
+      const transaction = await createNewTransaction(referenceNumber, { agreementId, id })
+      return { agreementId, transaction }
+    })
+  )
```
irisfaraway: Some of this feels like it might be a good candidate for abstracting out so it's not all sitting within `requestPayments()` as there's now quite a lot going on in here

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+    queueRecurringPayment(preparePayment(agreementId, transaction), batcher)
   }
+
+  await batcher.fetch()
+
+  logErrors(batcher.responses, 'Error requesting payments:')
+  await processPaymentResponses(paymentsToRequest, batcher)
+
+  return Promise.all(
+    batcher.responses.filter(r => r.status && !isClientError(r.status) && !isServerError(r.status)).map(async r => r.jsonValue)
```
irisfaraway: Can we abstract this filtering out into a more descriptive name?

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -47,11 +50,18 @@ const processRecurringPayments = async () => {
...
+  const requestedPayments = await requestPayments(dueRCPayments)
+  const payments = requestedPayments.map(payment => {
+    const duePayment = dueRCPayments.find(dueRCPayment => dueRCPayment.entity.agreementId === payment.agreement_id)
+    return {
+      paymentId: payment.payment_id,
+      agreementId: duePayment.entity.agreementId,
+      recurringPaymentId: duePayment.entity.id,
+      transactionId: payment.reference
+    }
+  })
```
irisfaraway: `processRecurringPayments()` is getting pretty long, so might be worth breaking this out into a separate `mapRequestedPayments()` or something

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -40,46 +40,79 @@ jest.mock('@defra-fish/connectors-lib', () => ({
...
+const getJSONImplementation = responseJSON => {
+  return function () {
+    if (!this.called) {
+      this.called = true
+      return Promise.resolve(responseJSON)
+    }
+    return Promise.reject(new TypeError('body used already'))
+  }
+}
+// have to go through all tests and check due payments are matching HTTP batcher responses
```
irisfaraway: Not sure what this is referring to - is this a todo note that accidentally got left in?

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -301,101 +330,208 @@ describe('recurring-payments-processor', () => {
...
+      // no need to mock returns for payment status batcher as we are only testing transaction creation errors here
 
       await execute()
 
       expect(console.error).toHaveBeenCalledWith(expect.any(String), ...errors)
     })
 
-    describe('when the error is caused by an invalid agreementId', () => {
-      it('logs out the ids', async () => {
+    it('logs an error for every failed payment', async () => {
```
irisfaraway: Might be worth testing with multiple errors also as right now we only test that one message is logged, and also don't test which payment it's being logged for

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -301,101 +330,208 @@ describe('recurring-payments-processor', () => {
...
         expect(console.log).toHaveBeenCalledWith(
           '%s is an invalid agreementId. Recurring payment %s will be cancelled',
-          'agreement-1',
-          'recurring-payment-1'
+          agreementId,
+          referenceNumber
         )
       })
 
       it('cancels the recurring payment', async () => {
```
irisfaraway: I'd like to see the invalid/inactive reason included in the test name, like it is for the test above - easier to read in the logs

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -301,101 +330,208 @@ describe('recurring-payments-processor', () => {
...
+        salesApi.getDueRecurringPayments.mockReturnValueOnce(Array(responses.length).fill(getMockDueRecurringPayment()))
+        HTTPRequestBatcher.mockImplementationOnce(getBatcherImplementation(responses))
 
-      await execute()
+        await execute()
 
-      expect(getPaymentStatus).toHaveBeenCalledTimes(6)
-    })
-  })
+        const [, paymentStatusBatcher] = HTTPRequestBatcher.mock.instances
```
irisfaraway: This comma deconstruction syntax wasn't familiar to me so I had to look it up! Wonder if `const paymentStatusBatcher = HTTPRequestBatcher.mock.instances[1]` might have been easier to read? I think that's more in line with what we've done elsewhere. Same goes for the next test down.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -745,64 +924,70 @@ describe('recurring-payments-processor', () => {
     })
 
     it('does not abort when getPaymentStatus rejects for one payment (allSettled at status stage)', async () => {
```
irisfaraway: Since `getPaymentStatus` is no longer being used here, this test may need renaming to clarify what is going on

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -40,46 +40,79 @@ jest.mock('@defra-fish/connectors-lib', () => ({
...
-        }
-      }
-    }
-  }
-]
-
-const getMockDueRecurringPayment = ({ agreementId = 'test-agreement-id', id = 'abc-123', referenceNumber = '123' } = {}) => ({
+const getPaymentStatusSuccess = (additionalFields = {}) => ({ state: { status: PAYMENT_STATUS.Success }, ...additionalFields })
+const getPaymentStatusFailure = () => ({ state: { status: PAYMENT_STATUS.Failure } })
+const getPaymentStatusError = () => ({ state: { status: PAYMENT_STATUS.Error } })
```
irisfaraway: I know these are still being used, but I wonder if they should be renamed for clarity since `getPaymentStatus` is no longer being used in this processor

irisfaraway (changes requested):
Nothing major, it's just a massive PR! There are some places where I'd maybe like to see things abstracted a bit further. Honestly I may take a second pass at reviewing the test suite just to confirm all different scenarios are covered, but submitting the review for now so GitHub doesn't lose it.

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -47,11 +50,18 @@ const processRecurringPayments = async () => {
     return
```
irisfaraway, line 49: This isn't a new issue to this PR but I don't think we have a test specifically checking that this return works. Maybe one that confirms that if there are no results, none of the subsequent calls to GOV.UK Pay are made?

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -265,31 +289,36 @@ describe('recurring-payments-processor', () => {
...
+      expect(queueRecurringPayment).toHaveBeenNthCalledWith(
         4,
-        expect.objectContaining({ ...expectedData, reference: 'transaction-id-4', agreement_id: agreementIds[3] })
+        expect.objectContaining({ ...expectedData, reference: 'transaction-id-4', agreement_id: agreementIds[3] }),
+        expect.anything()
       )
     })
 
-    it('logs an error for every failure', async () => {
+    it('logs an error for create transaction failures', async () => {
```
irisfaraway: The title of this test also specifies "create transaction failures" but it seems like it's testing errors from both `salesApi.preparePermissionDataForRenewal()` and `salesApi.createTransaction()`. I wonder if it's worth having separate tests for these, or at least updating the name of the test to clarify it is testing multiple types of failure. If we split it out then we could also test the content of the error message, rather than just `expect.any(String)`

Also, should we be checking that if either `salesApi.preparePermissionDataForRenewal()` or `salesApi.createTransaction()` fails for one record, it doesn't derail the process for subsequent records? We have some mocks here to ensure the others are still going through, but we aren't explicitly testing that they do.

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+            entity: { referenceNumber }
+          }
+        }
+      } = duePayment
+      const transaction = await createNewTransaction(referenceNumber, { agreementId, id })
+      return { agreementId, transaction }
+    })
+  )
+  logErrors(createTransactionResults, 'Error creating transactions:')
+  const paymentsToRequest = createTransactionResults.filter(ctr => ctr.status === 'fulfilled')
```
irisfaraway: I'm not sure we are explicitly testing this filtering anywhere.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -301,101 +330,208 @@ describe('recurring-payments-processor', () => {
...
-      expect(getPaymentStatus).toHaveBeenCalledTimes(6)
-    })
-  })
+        const [, paymentStatusBatcher] = HTTPRequestBatcher.mock.instances
+        paymentIds.forEach((paymentId, idx) => {
+          expect(queueRecurringPaymentStatusCheck).toHaveBeenNthCalledWith(idx + 1, paymentId, paymentStatusBatcher)
+        })
+      })
+
+      it('initiates the batcher to retrieve payment statuses', async () => {
```
irisfaraway, line 548: I wasn't able to find a corresponding test specifically for calling fetch against the payment request batcher (although we are testing that we add multiple items to its queue on line 1389). Think that might be good to have.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -301,101 +330,208 @@ describe('recurring-payments-processor', () => {
...
+        ],
+        statusResponses: [
+          { status: 200, json: getJSONImplementation(getPaymentStatusSuccess()) },
+          { status: 200, json: getJSONImplementation(getPaymentStatusSuccess()) },
+          { status: 200, json: getJSONImplementation(getPaymentStatusSuccess()) }
+        ]
+      })
+
+      await execute()
+      expect(console.error).toHaveBeenCalledWith(expect.stringMatching(/Unexpected response from GOV.UK Pay API\..*/))
```
irisfaraway, line 411: I'd like to see this test the "Error requesting payments:" message that gets passed into `logErrors()` as well, just so we can be clearer how the error is being called

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
-    return await sendPayment(preparedPayment)
-  } catch (error) {
-    if (error.message.includes('Invalid attribute value: agreement_id. Agreement does not exist')) {
-      console.log('%s is an invalid agreementId. Recurring payment %s will be cancelled', agreementId, transaction.recurringPayment.id)
-      await salesApi.cancelRecurringPayment(transaction.recurringPayment.id)
+const processPaymentResponses = async (paymentsToRequest, batcher) => {
+  for (let x = 0; x < paymentsToRequest.length; x++) {
+    const { value: paymentToRequest } = paymentsToRequest[x]
+    const response = batcher.responses[x]
+    response.jsonValue = response.json ? await response.json() : {}
```
irisfaraway, line 144: Can we add a test for the empty hash fallback here?

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+
+    if (isSuccessfulResponse(response.status)) {
+      await salesApi.createPaymentJournal(paymentToRequest.transaction.id, {
```
irisfaraway, line 147: I couldn't find any tests for what happens if `createPaymentJournal` throws an error

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -65,61 +75,94 @@ const fetchDueRecurringPayments = async date => {
...
+      const description = response.jsonValue.description || ''
+      if (
+        description.match(new RegExp(`Invalid attribute value: ${paymentToRequest.agreementId}. Agreement (does not exist|must be active)`))
+      ) {
+        console.log(
+          '%s is an invalid agreementId. Recurring payment %s will be cancelled',
+          paymentToRequest.agreementId,
+          paymentToRequest.transaction.recurringPayment.id
+        )
+        await salesApi.cancelRecurringPayment(paymentToRequest.transaction.recurringPayment.id)
```
irisfaraway, line 162: Do we need some testing for what happens if `salesApi.cancelRecurringPayment` raises an error?

packages/recurring-payments-job/src/recurring-payments-processor.js
irisfaraway: I also couldn't find any explicit tests for this filtering.

Reply (doing it here as it had marked the code as outdated):
It's tested in 'queues a payment status check for each successful payment response', using the second set of test cases where we have a 4xx and 5xx response.

irisfaraway (changes requested):
Second pass just to look for possible gaps in testing.

packages/connectors-lib/src/govuk-pay-api.js
jaucourt: Yes, it absolutely should. What I'm wondering about is how, when I tested it, a recurring payment was taken...

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
jaucourt: I'm not sure I fully understand, but I'll add a test that mocks out 429 responses below the maxRequestAttempts and then adds a 200 response, then checks that retries were only made until a successful response is received.

If the test didn't stop at maxRequestAttempts, this test would fail because the number of calls to `fetch` would exceed maxRequestAttempts, and the test wouldn't pass.

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
jaucourt: I would say that's done by 'calls fetch for each item in the queue' on line 64, but perhaps there's value in testing every response, so I'll amend the test to do that.

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: I was keeping this as it was in the main branch because it wasn't really included in the scope of the ticket, but it's no big deal to change it.

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: 1) Absolutely
2) They're in 'processRecurringPayments requests payment status for all %i successful payments when responses include %s', the second test case

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: This is a fair point, but given it isn't really within the scope of the ticket, I'll create a new ticket for the work (IWTF-5190)

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: I think they still describe what they are quite well, even though the getPaymentStatus function has been removed. Honestly can't think of a more descriptive name, let me know if you can.

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: I'll add this to the testing ticket I created

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: Updated the name on this - if you want to add to the testing ticket with the other requirements, then go ahead. We do test that preparePermissionDataForRenewal is called as we'd expect, so I don't have undue worries about this interface.

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: Great spot, added a test for this

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: Another good spot, added

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: This has changed now, as I realised logErrors wouldn't do anything with batcher results that failed, because they report as successful. The errors are actually handled in processPaymentResponse

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: Added ('lack of a json function on the error response is handled gracefully')

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: Added to 5190

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: Added to 5190

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -1321,6 +1438,130 @@ describe('recurring-payments-processor', () => {
...
+        { status: 200, reference: paymentIds[0], json: getJSONImplementation(getPaymentStatusSuccess({ payment_id: paymentIds[0] })) }
+      ]
+    })
+
+    await execute()
+
+    // console.log(queueRecurringPaymentStatusCheck.mock.calls)
+
+    // expect(queueRecurringPaymentStatusCheck).toHaveBeenNthCalledWith(1, paymentIds[0], expect.any(HTTPRequestBatcher))
+    // expect(queueRecurringPaymentStatusCheck).toHaveBeenNthCalledWith(2, paymentIds[1], expect.any(HTTPRequestBatcher))
```
irisfaraway: Some commented-out bits here

packages/connectors-lib/src/__tests__/govuk-pay-api.spec.js
irisfaraway: (Reply to an older conversation) Still wondering about this

packages/recurring-payments-job/README.md
```diff
@@ -10,7 +10,7 @@ When the RP job runs, all RP entries with an nextDueDate of the current date, a
 
 | name                           | description                                                     | required | default             | valid                         | notes |
 | ------------------------------ | --------------------------------------------------------------- | :------: | ------------------- | ----------------------------- | ----- |
-| NODE_ENV                       | Node environment                                                |    no    |                     | development, test, production |       |
+| NODE_ENV                       | Node environment                                                |    no    |                     | development, fsh-{env}.       |       |
```
irisfaraway: There's an extra full stop at the end here

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
```diff
@@ -190,6 +232,21 @@ describe('HTTP Request Batcher', () => {
       expect(fetch).toHaveBeenCalledTimes(maxRequestAttempts)
     })
 
+    // it.each([10, 7])('only retries until a successful response is received when maxRequestAttempts is set to %i', async maxRequestAttempts => {
```
irisfaraway: Commented-out code here. I'd probably prefer the commented-out version so we can test the variety, but if we're sticking with just testing 7, could probably simplify the test description so we don't iterate over an array with only one entry

packages/recurring-payments-job/src/recurring-payments-processor.js
irisfaraway: (Reply to an older conversation) I can see from the commit history that we are testing that ones that throw errors are not created, but still doesn't look like we are testing against the status being 'fulfilled' - am I missing something?

PB: Can't reply, so editing your comment again! This is tested by 'only creates payments for transactions that are successfully created'.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
irisfaraway: (Reply to an older conversation) I was trying to think up names that didn't feel like they could be confused with `getPaymentStatus()`, but then it occurred to me that with `getPaymentStatus()` removed from the processor, we actually aren't using it anywhere else. So maybe just rip that function out completely and leave these mocks as they are instead?

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -366,11 +407,12 @@ describe('recurring-payments-processor', () => {
       })
 
       await execute()
+
       expect(console.error).toHaveBeenCalledWith(expect.stringMatching(/Unexpected response from GOV.UK Pay API\..*/))
```
irisfaraway, line 411: Is it worth testing that the full error data is logged correctly (with the error message, stringified payload, etc?)

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -343,20 +382,22 @@ describe('recurring-payments-processor', () => {
...
+    it('logs an error when a payment fails', async () => {
       jest.spyOn(console, 'error')
       salesApi.getDueRecurringPayments.mockReturnValueOnce([
         getMockDueRecurringPayment({ referenceNumber: 'fee', agreementId: 'a1' }),
         getMockDueRecurringPayment({ referenceNumber: 'fi', agreementId: 'a2' }),
         getMockDueRecurringPayment({ referenceNumber: 'foe', agreementId: 'a3' }),
         getMockDueRecurringPayment({ referenceNumber: 'fum', agreementId: 'a4' })
       ])
+      const errorResponse = new Error('Payment request failed')
+      errorResponse.reference = 'a2'
```
irisfaraway: I'm not sure where we do anything with this?

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -574,6 +630,15 @@ describe('recurring-payments-processor', () => {
     expect(salesApi.createTransaction).toHaveBeenCalledWith(expectedData)
   })
 
+  it('only creates payments for transactions that are successfully created', async () => {
+    salesApi.getDueRecurringPayments.mockReturnValueOnce([getMockDueRecurringPayment(), getMockDueRecurringPayment()])
+    salesApi.createTransaction.mockRejectedValueOnce(new Error('Transaction creation failed'))
+
+    await execute()
+
+    expect(queueRecurringPayment).toHaveBeenCalledTimes(1)
```
irisfaraway, line 639: Would prefer to test what we're calling it with, rather than just the number of calls

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -1321,6 +1438,130 @@ describe('recurring-payments-processor', () => {
...
+            created_date: ''
+          })
+        }
+      ]
+    })
+    salesApi.getPaymentJournal.mockResolvedValueOnce(true)
+
+    await execute()
+
+    expect(salesApi.processRPResult).toHaveBeenCalledWith('test-transaction-id', 'test-payment-id', expect.any(String))
```
irisfaraway: I'm not sure this succeeds in testing the 429 for the first batcher, since we are mocking the second batcher to always return a success response with those payment identifiers. At the moment I think we could swap out `paymentResponse` with totally different values and this would still pass.

Maybe if we throw the 429 into `statusResponses` instead, or test that the second batcher is calling the values provided? Although that may already be covered elsewhere. I don't know, it's an awkward one.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -1321,6 +1438,130 @@ describe('recurring-payments-processor', () => {
...
+      .mockReturnValueOnce({
+        cost: 44,
+        id: 'transaction-2'
+      })
+    const paymentResponses = [
+      { status: 200, reference: agreementIds[1], json: getJSONImplementation({ payment_id: 'payment-2', agreement_id: agreementIds[1] }) },
+      { status: 200, reference: agreementIds[0], json: getJSONImplementation({ payment_id: 'payment-1', agreement_id: agreementIds[0] }) }
+    ]
+    mockBatchers({
+      // payment responses are intentionally out of order to test matching by agreementId
```
irisfaraway: Can we move this comment up to where `paymentResponses` are defined? Would be slightly clearer

irisfaraway (changes requested):
Refactor looks good. I'm definitely happier with how the matching works now :) No major feedback, just a few suggestions around testing and some commented-out lines. I did also reply to some of my older comments so hopefully that comes through okay (I think we can rip out `getPaymentStatus()` entirely!)

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: I'm not a fan of testing the full error message, at least in this case, because I think it makes for a brittle test

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: We do not - I'll delete

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: We do that elsewhere, though - this is literally just to test that queueRecurringPayment isn't called for a transaction failure

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: I've refactored this slightly so we're using proper guids for the various identifiers, but mocking out the first batcher is what we want - we mimic a 429 response, which the real batcher would retry, so I model that with a subsequent 200 response. The expected behaviour is that the recurring payments processor takes the last (successful) response from the array of responses and then continues with making a payment status response and checks the record is processed. I suppose we could change to simply making sure that the payment status request is done and ignoring the call to processRPResult, but the effect would be the same.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
jaucourt: Done

packages/recurring-payments-job/src/recurring-payments-processor.js
jaucourt: I've thought about this further, and I'd rather add it to 5190 if we need to. Honestly, I think that whether the agreementId doesn't exist or is inactive - and the former will rarely if ever happen - can be treated as one and the same by this code. If we want to investigate further we can check the agreement id manually, but I doubt that we would, we'd likely just leave it as cancelled.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
irisfaraway: Ahh gotcha. LGTM then!

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
irisfaraway: Yeah that's fair, I can't really think of a better way to handle it!

packages/connectors-lib/src/__tests__/http-request-batcher.spec.js
jaucourt: Nope, my mistake, I've put both versions back in

irisfaraway (approved):
🎺 🎺 🎺 :shipit: 🎉

PR #2136, opened 2025-03-12
Ability to turn off rcp

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4547

Adjust so when turn off rcp wont have any impact

### DIFF
--- a/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
+++ b/packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
@@ -50,20 +50,15 @@ describe('validForRecurringPayment', () => {
 
 describe('isRecurringPayment', () => {
   it.each`
-    show     | agreementId  | expected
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/processors/__tests__/recurring-pay-helper.spec.js
```diff
@@ -53,7 +53,7 @@ describe('isRecurringPayment', () => {
     show     | agreementId  | expected
     ${true}  | ${'foo123'}  | ${true}
     ${true}  | ${undefined} | ${false}
-    ${false} | ${'foo123'}  | ${false}
+    ${false} | ${'foo123'}  | ${true}
```
irisfaraway, line 56: I don't think we need to be testing the value of SHOW_RECURRING_PAYMENTS at all if it no longer has any bearing on the code, so reckon you can simplify this test a bit

irisfaraway (changes requested):

irisfaraway (approved):

PR #2138, opened 2025-03-25
Implement fail screen for easy renewals journey Part 1

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4392

The easy renewals journey is using error validation to tell a user they are not eligible to use the service (in this case because their details do not match any in our systems). This is not a correct use of the design pattern: [Recover from validation errors](https://design-system.service.gov.uk/patterns/validation/)

“Do not use validation to check whether the user is eligible to use the service”

Instead the user should be taken to a new screen that tells them what has happened and what they need to do next.


packages/gafl-webapp-service/src/pages/renewals/identify/__tests__/identify.spec.js
```diff
@@ -52,14 +53,11 @@ mockSalesApi()
...
-  it('redirects to identify page when called with an invalid permission reference', async () => {
+  it('redirects from identify to licence not found page when called with an invalid permission reference', async () => {
     const data = await injectWithCookies('GET', RENEWAL_PUBLIC.uri.replace('{referenceNumber}', 'not-a-valid-reference-number'))
     expect(data.statusCode).toBe(302)
     expect(data.headers.location).toHaveValidPathFor(IDENTIFY.uri)
-    const data2 = await injectWithCookies('GET', IDENTIFY.uri)
-    expect(data2.statusCode).toBe(302)
-    expect(data2.headers.location).toHaveValidPathFor(IDENTIFY.uri)
-    const data3 = await injectWithCookies('GET', IDENTIFY.uri)
+    const data3 = await injectWithCookies('GET', LICENCE_NOT_FOUND.uri)
```
irisfaraway, line 60: This can probably just be `data2` again if that no longer exists

packages/gafl-webapp-service/src/pages/renewals/identify/__tests__/identify.spec.js
```diff
@@ -87,7 +85,7 @@ describe('The easy renewal identification page', () => {
     expect(data.headers.location).toHaveValidPathFor(IDENTIFY.uri)
   })
 
-  it('redirects back to itself on posting an valid but not authenticated details', async () => {
+  it('redirects to licence not found on posting an valid but not authenticated details', async () => {
```
irisfaraway, line 88: Not a change you made but "an valid" is bugging me!
```suggestion
  it('redirects to licence not found on posting valid but not authenticated details', async () => {
```

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
```diff
@@ -15,11 +15,7 @@
     set errorMap = {
         'referenceNumber' : {
             'string.empty': { text: mssgs.identify_error_empty, ref: '#ref' },
-            'string.pattern.base': { text: mssgs.identify_error_pattern, ref: '#ref' },
-            'string.invalid': {
-                text: mssgs.identify_error_invalid_1 + payload.referenceNumber + mssgs.identify_error_invalid_2,
-                ref: "#ref"
-            }
+            'string.pattern.base': { text: mssgs.identify_error_pattern, ref: '#ref' }
```
irisfaraway, line 18: If `mssgs.identify_error_invalid_1` and `mssgs.identify_error_invalid_2` are no longer used anywhere, can we take them out of the lang files?

packages/gafl-webapp-service/src/routes/journey-definition.js
```diff
@@ -413,7 +414,13 @@ export default [
       [CommonResults.OK]: {
         page: LICENCE_SUMMARY
       }
-    }
+    },
+    backLink: IDENTIFY.uri
```
irisfaraway, line 418: Just wondering why this was added to modify the identify page - was it previously a broken link?

packages/gafl-webapp-service/src/handlers/authentication-handler.js
```diff
@@ -42,10 +42,7 @@ export default async (request, h) => {
   }
 
   if (!authenticationResult) {
-    payload.referenceNumber = referenceNumber
-    await request.cache().helpers.page.setCurrentPermission(IDENTIFY.page, { payload, error: { referenceNumber: 'string.invalid' } })
-    await request.cache().helpers.status.setCurrentPermission({ referenceNumber, authentication: { authorized: false } })
```
irisfaraway, line 47: Do we need to be setting this info anywhere else now that it's been removed? Not sure of the particulars, but it's possible we have other reasons to want to save this, for example hanging on to the reference number when the user goes back to the identify page from the error page.

irisfaraway (changes requested):
This is looking broadly good, just a few things to tidy up and a few questions from me :)

Also I noticed that the comments on lines 10-14 of the authentication handler can probably be updated now that we display the errors on another page.

packages/gafl-webapp-service/src/locales/en.json
```diff
@@ -415,6 +415,14 @@
   "licence_length_one_day": "1 day licences are valid for 24 hours from the start time that you select.",
   "licence_length_title_other": "How long do they want the licence for?",
   "licence_length_title_you": "How long do you want your licence for?",
+  "licence_not_found_title": "Licence cannot be found ",
```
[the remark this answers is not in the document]
lailien3: Screenshot 2025-04-17 at 12 23 23
Yes, it is used in the tab name to ensure there is a gap between the title and the service name :)

packages/gafl-webapp-service/src/locales/cy.json
```diff
@@ -416,6 +416,14 @@
   "licence_length_title_other": "Am ba hyd y maent am gael y drwydded?",
   "licence_length_title_you": "Am ba hyd yr hoffech chi gael y drwydded?",
   "licence_num": "Rhif trwydded",
+  "licence_not_found_title": "Methu dod o hyd i drwydded ",
```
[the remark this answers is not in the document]
lailien3: Yes, same reason as for the english :) (needed to add the gap between the title and service name in the tab)
Screenshot 2025-04-17 at 12 24 59

packages/gafl-webapp-service/src/pages/renewals/identify/__tests__/identify.spec.js
lailien3: good spot!

packages/gafl-webapp-service/src/routes/journey-definition.js
lailien3: omg ty for noticing! added this very early on and forgot to remove it!

packages/gafl-webapp-service/src/pages/renewals/identify/identify.njk
lailien3: checked & can confirm they are no longer used elsewhere; removed :)

packages/gafl-webapp-service/src/handlers/authentication-handler.js
lailien3: There seems to already be good data retention when going back and forth between the error pages and the identify page, but to err on the side of caution I ill put them back in :)

packages/gafl-webapp-service/src/pages/renewals/licence-not-found/route.js
```diff
@@ -0,0 +1,4 @@
+import { LICENCE_NOT_FOUND } from '../../../uri.js'
+import pageRoute from '../../../routes/page-route.js'
+
+export default pageRoute(LICENCE_NOT_FOUND.page, LICENCE_NOT_FOUND.uri, null)
```
jaucourt, line 4: Strictly speaking, there should be tests for this. If there are, elsewhere, then they should really be in `licence-not-found/__tests__`

jaucourt (changes requested):

packages/gafl-webapp-service/src/pages/renewals/licence-not-found/route.js
lailien3: np, it was sat in identify/route.spec :) ive moved it here

packages/gafl-webapp-service/src/pages/renewals/licence-not-found/__tests__/route.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    )
+    expect(data.statusCode).toBe(302)
+    expect(data.headers.location).toHaveValidPathFor(AUTHENTICATE.uri)
+    const data2 = await injectWithCookies('GET', AUTHENTICATE.uri)
+    expect(data2.statusCode).toBe(302)
+    expect(data2.headers.location).toHaveValidPathFor(LICENCE_NOT_FOUND.uri)
+    const data3 = await injectWithCookies('GET', LICENCE_NOT_FOUND.uri)
+    expect(data3.statusCode).toBe(200)
+  })
+})
```
jaucourt, line 32: Right. I've never liked these tests, and I never will. They're not a unit test and they're not an acceptance test.

What you need to replace it with is:
1) a test to check that the first argument passed to pageRoute is LICENCE_NOT_FOUND.page
2) a test to check that the second argument passed to pageRoute is LICENCE_NOT_FOUND.uri

To do this you'll need to mock pageRoute and the LICENCE_NOT_FOUND variable; send through a symbol for each argument. As for the third argument, I'm fairly sure there's no need to pass a null value, you can simply omit it.

lailien3:
> Looks lovely now. Sorry if my last comment sounded a bit narky, nothing personal I simply hate those injectWithCookies tests and forgot my manners!

That's alright!! I recycled the test without too much thought into it (definitely much cleaner now 🤣)

packages/gafl-webapp-service/src/locales/cy.json
```diff
@@ -600,7 +606,6 @@
   "payment_cancelled_title": "Mae eich taliad wedi cael ei ganslo",
   "payment_failed_expired": "Ni chafodd y taliad ei gwblhau o fewn 90 munud o greu’r cais",
   "payment_failed_not_taken": "Nid oes unrhyw arian wedi cael ei dynnu o’ch cyfrif.",
-  "payment_failed_not_taken_recurring": "Nid oes unrhyw arian wedi cael ei dynnu o’ch cyfrif ac nid yw eich manylion talu wedi cael eu harbed ar gyfer y taliad cerdyn sy’n ailadrodd.",
```
irisfaraway, line 603: Why is this being removed? And only from the Welsh but not the English?

lailien3: I have no idea why, its deffo still there locally 🤔

irisfaraway (approved):

PR #2139, opened 2025-04-03
Card digits dynamics lib

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4610

Add last card digits field into dynamics lib


irisfaraway (approved):

packages/dynamics-lib/src/entities/recurring-payment.entity.js
```diff
@@ -24,7 +24,8 @@ export class RecurringPayment extends BaseEntity {
       activePermission: { field: '_defra_activepermission_value', type: 'string' },
       contactId: { field: '_defra_contact_value', type: 'string' },
       publicId: { field: 'defra_publicid', type: 'string' },
-      nextRecurringPayment: { field: '_defra_nextrecurringpayment_value', type: 'string' }
+      nextRecurringPayment: { field: '_defra_nextrecurringpayment_value', type: 'string' },
+      lastDigitsCardNumber: { field: 'defra_last_digits_card_number', type: 'integer' }
```
jaucourt, line 28: Not at all sure about this being an integer - what about leading zeros in the card number? Would have expected it to be a string

jaucourt (commented):
Think this might need to be amended - and could need Ravi to make changes too.

PR #2140, opened 2025-04-04
Upgrade AWS SDK to v3

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-3449

Upgrading AWS SDK to v3

When making this upgrade, I have tried to keep clients of connectors-lib as close to the original implementation as possible - mostly there are spacing changes and removal of the `.promise()` suffix as it's no longer needed in AWS SDK v3. There are more significant changes in tests - I've generally kept these the same, with a few exceptions where I've simplified and streamlined tests to remove multiple expect statements. I've removed the separate mocking of AWS in favour of mocks kept in the same file as the tests, as this makes it more obvious what's being done and avoids overly complex mocks.

The biggest changes are in the connectors lib, as this is where the instantiation occurs so changes were necessary. I've still done my best to keep changes to a minimum, mostly it's multiple imports as there isn't one monolithic AWS SDK anymore.

Finally, despite the ticket it's attached to, it's an upgrade of the whole SDK, not just the DynamoDB part. It was easier to do the whole lot rather than pick out individual parts.


nabeelamir-defra (approved):

irisfaraway (commented):
What a monster! Code looks good to me but appreciate that with all the lib changes we can't fully test this until merging.

Sonar does say there's a few lines missing test coverage though - I think in the Sales API health check.

irisfaraway (approved):

PR #2143, opened 2025-04-14
Link old and new recurring payments after payment

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-4513

When we take a recurring payment, we want to create a new recurring payment record and update the existing record with a reference to the new one.


packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
```diff
@@ -410,6 +418,35 @@ describe('transaction service', () => {
...
+        AwsMock.DynamoDB.DocumentClient.__setResponse('get', { Item: finalisedTransaction })
+        findNewestExistingRecurringPaymentInCrm.mockReturnValueOnce(mockExistingRecurringPayment)
+
+        await processQueue({ id: finalisedTransaction.id })
+
+        expect(mockExistingRecurringPayment.bindToEntity).toHaveBeenCalledWith(
+          RecurringPayment.definition.relationships.nextRecurringPayment,
+          mockRecurringPayment
+        )
+      })
```
jaucourt, line 449: You're not checking the negative here - i.e. if findNewestExistingRecurringPaymentInCrm returns `false` that the result doesn't contain an existingRecurringPayment

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -126,3 +126,12 @@ export const processRPResult = async (transactionId, paymentId, createdDate) =>
 
   return { permission }
 }
+
+export const findNewestExistingRecurringPaymentInCrm = async agreementId => {
+  const matchingRecords = await executeQuery(findRecurringPaymentsByAgreementId(agreementId))
+  if (matchingRecords) {
+    return [...matchingRecords].sort((a, b) => a.endDate.localeCompare(b.endDate)).pop()
+  } else {
+    return false
```
jaucourt, line 135: Doesn't look like you're testing for this - you test for returnedRecord to be `undefined` but not a boolean `false`

jaucourt (commented):
Leaving my remarks as comments rather than explicit approval / changes requested so that this can be merged in my absence

packages/sales-api-service/src/services/transactions/__tests__/process-transaction-queue.spec.js
```diff
@@ -410,6 +418,57 @@ describe('transaction service', () => {
...
+        const agreementId = Symbol('agreementId')
+        const mockRecurringPayment = { agreementId }
+        const finalisedTransaction = mockFinalisedTransactionRecord()
+        processRecurringPayment.mockResolvedValueOnce({ recurringPayment: mockRecurringPayment })
+        AwsMock.DynamoDB.DocumentClient.__setResponse('get', { Item: finalisedTransaction })
+        findNewestExistingRecurringPaymentInCrm.mockReturnValueOnce(false)
+
+        await processQueue({ id: finalisedTransaction.id })
+
+        expect(persist.mock.calls[0][0]).toEqual([
```
ScottDormand96, line 461: don't think should be so many expect statements, maybe could look into using snapshot

ScottDormand96 (changes requested):

ScottDormand96 (approved):

PR #2147, opened 2025-04-28
RP job handles API errors

lailien3:

Adding logging and aborting run mechanism

### DIFF
--- a/packages/recurring-payments-job/src/recurring-payments-processor.js
+++ b/packages/recurring-payments-job/src/recurring-payments-processor.js
@@ -11,12 +11,17 @@ export const processRecurringPayments = async () => {
   if (process.env.RUN_RECURRING_PAYMENTS?.toLowerCase() === 'true') {
     console.log('Recurring Payments job enabled')
     const date = new Date().toISOString().split('T')[0]
-    const response = await salesApi.getDueRecurringPayments(date)
-    console.log('Recurring Payments found: ', response)
-    await Promise.all(response.map(record => processRecurringPayment(record)))
-    if (response.length > 0) {
-      await new Promise(resolve => setTimeout(resolve, PAYMENT_STATUS_DELAY))
-      await Promise.all(response.map(record => processRecurringPaymentStatus(record)))
+    try {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -104,6 +104,10 @@
     ],
     "testEnvironment": "node",
     "testRunner": "jest-circus/runner",
-    "silent": true
+    "silent": false,
+    "verbose": true
```
ScottDormand96, line 108: you must've forgot change this back

package.json
```diff
@@ -104,6 +104,10 @@
...
+  },
+  "dependencies": {
+    "http-response-status-code": "^1.7.5"
```
ScottDormand96, line 111: should this be moved too recurring-payments package.json?

packages/recurring-payments-job/src/recurring-payments-processor.js
```diff
@@ -101,13 +107,28 @@ const preparePayment = (agreementId, transaction) => {
...
+    const {
+      state: { status }
+    } = await getPaymentStatus(paymentId)
+
+    console.log(`Payment status for ${paymentId}: ${status}`)
+    if (status === PAYMENT_STATUS_SUCCESS) {
+      const payment = payments.find(p => p.paymentId === paymentId)
+      await salesApi.processRPResult(payment.transaction.id, paymentId, payment.created_date)
+    }
+  } catch (error) {
```
ScottDormand96, line 121: just a question here - if it isn't successful is it an error? or is there other scenarios?

ScottDormand96 (changes requested):

package.json
lailien3: 🫡

package.json
lailien3: yes

jaucourt (changes requested):
Some rework on the handling of RCP errors to make the error more specific. However, you're missing handling for the third AC: "If systems are down mid-run, continue with the run and try again": i.e. request further payments.
Also, you'll need to check that when errors are logged, they output to errbit and Slack so that we're properly notified (unless you've already done this).

PR #2152, opened 2025-05-13
Remove date picker from start kind screen

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-4479

There are a number of problems with the date picker: it's not accessible for those using assistive technology, it's not multi-lingual, it relies on Javascript and finally, it was never solving an identified user need.


lailien3 (approved):
LGTM :)

PR #2169, opened 2025-06-10
RP job handles GOV.UK Pay error responses

lailien3:

When the recurring payments job sends a “create payment” request to GOV.UK Pay, we expect it to return confirmation that our payment request has been received. However, if our request to the payment API returns an error response code (for example, if the API is down or if our authorisation is invalid) then the recurring payment job needs to handle this gracefully.

### DIFF
--- a/packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
+++ b/packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
@@ -1,7 +1,6 @@
 import { salesApi } from '@defra-fish/connectors-lib'
 import { processRecurringPayments } from '../recurring-payments-processor.js'
 import { getPaymentStatus, sendPayment } from '../services/govuk-pay-service.js'
-import db from 'debug'
 
 jest.mock('@defra-fish/business-rules-lib')
 jest.mock('@defra-fish/connectors-lib', () => ({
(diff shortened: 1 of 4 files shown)


lailien3:
look to https://github.com/DEFRA/rod-licensing/pull/2147 for review comments 😅

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -68,10 +82,149 @@ describe('recurring-payments-processor', () => {
...
+        2,
+        expect.objectContaining({ ...expectedData, reference: 'transaction-id-2', agreement_id: agreementIds[1] })
+      )
+      expect(sendPayment).toHaveBeenNthCalledWith(
+        3,
+        expect.objectContaining({ ...expectedData, reference: 'transaction-id-3', agreement_id: agreementIds[2] })
+      )
+      expect(sendPayment).toHaveBeenNthCalledWith(
+        4,
+        expect.objectContaining({ ...expectedData, reference: 'transaction-id-4', agreement_id: agreementIds[3] })
```
irisfaraway, line 183: This test is quite complex and has a lot of different assertions. Would it be worth splitting it out into multiple tests and refactoring some of the setup into a different function?

irisfaraway (changes requested):

PR #2179, opened 2025-07-17
Set name of RecurringPayment record

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-4701

Currently the name field of RecurringPayments in the CRM is not set to anything useful. We want it to include the contact name and the year the next payment is due to make it easier to find records in the interface.

### DIFF
--- a/packages/sales-api-service/src/services/recurring-payments.service.js
+++ b/packages/sales-api-service/src/services/recurring-payments.service.js
@@ -150,8 +150,6 @@ export const cancelRecurringPayment = async id => {
 }
 
 const determineRecurringPaymentName = (transactionRecord, contact) => {
-  console.log(transactionRecord)
-  console.log(contact)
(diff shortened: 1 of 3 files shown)


packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -148,3 +148,8 @@ export const cancelRecurringPayment = async id => {
     console.log('No matches found for cancellation')
   }
 }
+
+const determineRecurringPaymentName = (transactionRecord, contact) => {
+  const dueYear = transactionRecord.payment.recurring.nextDueDate.split('-')[0]
```
jaucourt, line 153: This is really nit-picky, but IMHO it'd be slightly neater to do `const [dueYear] = transactionRecord.payment.recurring.nextDueDate.split('-')`

jaucourt (changes requested):

jaucourt (approved):

PR #2183, opened 2025-07-23
RCP - Clean up from previous jobs

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-4667
    
The RCP job should pick up any 'hanging' transactions where the status didn't resolve in time on the previous run, re-checking the status and either processing successful payments or cancelling ones that still haven't resolved.


packages/payment-mop-up-job/src/processors/processor.js
```diff
@@ -18,6 +18,34 @@ const limiter = new Bottleneck({
...
+  const { code, state } = transaction.paymentStatus
+  const { paymentTimestamp } = transaction
+
+  const isExpiredCancelledOrRejected = [
+    GOVUK_PAY_ERROR_STATUS_CODES.EXPIRED,
+    GOVUK_PAY_ERROR_STATUS_CODES.USER_CANCELLED,
+    GOVUK_PAY_ERROR_STATUS_CODES.REJECTED
+  ].includes(state?.code)
+
+  const isExpiredNotFound =
```
irisfaraway, line 43: Can we keep the old `// The payment's not found and three hours have elapsed` comment here? Wasn't entirely sure what circumstances this was for at first

packages/payment-mop-up-job/src/processors/processor.js
```diff
@@ -18,6 +18,34 @@ const limiter = new Bottleneck({
...
+    case GOVUK_PAY_ERROR_STATUS_CODES.NOT_FOUND:
+      return PAYMENT_JOURNAL_STATUS_CODES.Expired
+    case GOVUK_PAY_ERROR_STATUS_CODES.USER_CANCELLED:
+      return PAYMENT_JOURNAL_STATUS_CODES.Cancelled
+    case GOVUK_PAY_ERROR_STATUS_CODES.REJECTED:
+      return PAYMENT_JOURNAL_STATUS_CODES.Failed
+  }
+}
+
+const shouldUpdatePaymentJournal = transaction => {
```
irisfaraway, line 33: I think it'd be worth clarifying the name of this function to say what specific conditions it's for, since there are several other circumstances that we update the payment journal anwyay. At the moment we update the payment journal if it's an error, we update the payment journal if it's a success, and only then do we check `shouldUpdatePaymentJournal`. Threw me off a bit initially.

packages/connectors-lib/src/__tests__/document-client-decorator.spec.js
```diff
@@ -1,6 +1,6 @@
 import { createDocumentClient } from '../documentclient-decorator'
 import { DynamoDB } from '@aws-sdk/client-dynamodb'
-import { DynamoDBDocument } from '@aws-sdk/lib-dynamodb'
+import { DynamoDBDocument, QueryCommand, ScanCommand } from '@aws-sdk/lib-dynamodb'
```
irisfaraway, line 3: Was this change needed to make this change work or was it just some refactoring? (Not saying we shouldn't make it, I'm just curious!)

irisfaraway (approved):
Much clearer! :)

PR #2184, opened 2025-07-28
Incorrect capitalistion of address in contact summary

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4719

On the licence summary page, the address all appears in capital letters, while it should only have the first letter capitalised. (Postcode and
Country)

### DIFF
--- a/packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
+++ b/packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
@@ -122,13 +122,13 @@ export class RowGenerator {
     }
 
     const formattedParts = [
-      licensee.premises && titleCase(licensee.premises),
-      licensee.street && titleCase(licensee.street),
-      licensee.locality && titleCase(licensee.locality),
-      licensee.town && titleCase(licensee.town),
-      licensee.postcode && licensee.postcode.toUpperCase(),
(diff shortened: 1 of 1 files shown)


packages/gafl-webapp-service/src/pages/summary/contact-summary/route.js
```diff
@@ -112,9 +112,29 @@ export class RowGenerator {
...
+    const formattedParts = [
+      titleCase(licensee.premises),
+      titleCase(licensee.street),
+      titleCase(licensee.locality),
+      titleCase(licensee.town),
+      licensee.postcode.toUpperCase(),
+      countryName.toUpperCase()
+    ]
+
+    const text = formattedParts.filter(Boolean).join(', ')
```
jaucourt, line 137: Honestly, this seems overly complex. What I think you should do is `toLowerCase()` on premises, street, locality and town, then in the browser set a style on that element of `text-transform: uppercase`. That's what we do in address lookup service when we format it for display when the user is selecting an address: https://github.com/DEFRA/rod-licensing/blob/435e74ebb124eeda4a409f5aa256ba188125f06b/packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js#L41

jaucourt (approved):

PR #2197, opened 2025-08-15
Recurring payment job should handle errors from await salesApi.processRPResult()

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4649

Currently, when the recurring payment job calls await salesApi.processRPResult, we await the response but we don’t do anything with it. We should have some error handling in place so that if one record fails to process, we can log it to be reviewed later, but also not have the whole job fall over.


jaucourt (dismissed):
I think you and I need to do a pairing session and we can work on refactoring the tests here to tighten things up.

PR #2200, opened 2025-08-27
Junior licence format

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4789

Update how we set name of junior licence, currently setting based on number and not comparing dates so rounding issue for 16/17 year olds

### DIFF
--- a/packages/sales-api-service/src/services/permissions.service.js
+++ b/packages/sales-api-service/src/services/permissions.service.js
@@ -95,8 +95,9 @@ const getAgeCategory = (birthDate, issueDate) => {
     return 'J'
   } else if (issue.isSameOrAfter(sixtysixthBirthday)) {
     return 'S'
+  } else {
+    return 'F'
   }
-  return 'F'
 }
 
 /**


packages/sales-api-service/src/services/permissions.service.js
```diff
@@ -86,8 +88,9 @@ const getAgeCategory = (birthDate, issueDate) => {
   const dob = moment(birthDate)
   const issue = moment(issueDate)
   const diff = issue.diff(dob, 'years', true)
+  const seventeenthBirthday = dob.clone().add(ADULT_AGE, 'years')
 
-  if (isJunior(diff)) {
+  if (issue.isBefore(seventeenthBirthday)) {
     return 'J'
   } else if (isSenior(diff)) {
```
irisfaraway, line 95: Just wondering, if using the diff like this was an issue for junior licences, is it also an issue for seniors?

I think it'd be worth having unit tests for birthdays which are in this rounding-up error zone.

irisfaraway (approved):

PR #2201, opened 2025-08-27
Create cancellation journey

jaucourt:
https://eaflood.atlassian.net/browse/IWTF-4773

Create the bare bones of the cancellation journey, along with a feature switch to disable the pages


packages/gafl-webapp-service/src/routes/telesales-routes.js
```diff
@@ -26,3 +30,11 @@ export default [
     options: { auth: false }
   }
 ]
+
```
ScottDormand96, line 33: shall we create a ticket to update this in future?

packages/gafl-webapp-service/src/routes/journey-definition.js
```diff
@@ -426,5 +430,31 @@ export default [
   {
     current: RENEWAL_START_DATE,
     backLink: LICENCE_SUMMARY.uri
+  },
+
+  // cancel RP journey
+  {
```
ScottDormand96, line 436: are back links required at this stage?

jaucourt: It wasn't mentioned in the ACs...

PR #2215, opened 2025-09-25
Recurring payment job should handle errors from await salesApi.processRPResult()

lailien3:
https://eaflood.atlassian.net/browse/IWTF-4649

Currently, when the recurring payment job calls await salesApi.processRPResult, we await the response but we don’t do anything with it. We should have some error handling in place so that if one record fails to process, we can log it to be reviewed later, but also not have the whole job fall over.

### DIFF
--- a/packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
+++ b/packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
@@ -70,8 +70,25 @@ const getMockDueRecurringPayment = ({ agreementId = 'test-agreement-id', id = 'a
   expanded: { activePermission: { entity: { referenceNumber } } }
 })
 
+const makeSendPaymentResponse = (overrides = {}) => ({
+  payment_id: 'pay-1',
+  created_date: '2025-01-01T00:00:00.000Z',
+  ...overrides
+})
+
 describe('recurring-payments-processor', () => {
   const [{ value: debugLogger }] = db.mock.results
+
+  const date1 = '2025-01-01T00:00:00.000Z'
+  const date2 = '2025-01-01T00:01:00.000Z'
+
+  const mockTwoDuePayments = () => {
+    salesApi.getDueRecurringPayments.mockResolvedValueOnce([getMockDueRecurringPayment(), getMockDueRecurringPayment()])
+
+    salesApi.preparePermissionDataForRenewal.mockResolvedValue({ licensee: { countryCode: 'GB-ENG' } })
+
+    salesApi.createTransaction.mockResolvedValueOnce({ id: 'trans-1', cost: 30 }).mockResolvedValueOnce({ id: 'trans-2', cost: 30 })
+  }
 
   beforeEach(() => {
     jest.clearAllMocks()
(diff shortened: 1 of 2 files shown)


packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -70,9 +70,26 @@ const getMockDueRecurringPayment = ({ agreementId = 'test-agreement-id', id = 'a
...
 describe('recurring-payments-processor', () => {
   const [{ value: debugLogger }] = db.mock.results
 
+  const date1 = '2025-01-01T00:00:00.000Z'
+  const date2 = '2025-01-01T00:01:00.000Z'
+
+  const mockTwoDuePayments = () => {
+    salesApi.getDueRecurringPayments.mockResolvedValueOnce([getMockDueRecurringPayment(), getMockDueRecurringPayment()])
+
+    salesApi.preparePermissionDataForRenewal.mockResolvedValue({ licensee: { countryCode: 'GB-ENG' } })
```
irisfaraway, line 88: Best to do `mockResolvedValueOnce` over `mockResolvedValue` so you don't have to worry about mocks spilling over to other tests.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -70,9 +70,26 @@ const getMockDueRecurringPayment = ({ agreementId = 'test-agreement-id', id = 'a
...
+  ...overrides
+})
+
 describe('recurring-payments-processor', () => {
   const [{ value: debugLogger }] = db.mock.results
 
+  const date1 = '2025-01-01T00:00:00.000Z'
+  const date2 = '2025-01-01T00:01:00.000Z'
+
+  const mockTwoDuePayments = () => {
```
irisfaraway, line 85: Rather than setting this up as a separate function, I would suggest putting it in as setup for a `before` block. So you could nest another `describe('when multiple payments are due')` block within `describe('recurring-payments-processor')`, move your tests that need this mocking into that, and then do a `beforeEach` to set up the mocking. I think that'd be more consistent with how we've done our setup for tests elsewhere in the service.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -588,6 +605,112 @@ describe('recurring-payments-processor', () => {
...
+    sendPayment.mockResolvedValueOnce(payment)
+
+    getPaymentStatus.mockResolvedValueOnce(getPaymentStatusSuccess())
+    salesApi.processRPResult.mockRejectedValueOnce(new Error('boom'))
+
+    const errorSpy = jest.spyOn(console, 'error').mockImplementation(() => {})
+
+    await execute()
+
+    expect(errorSpy).toHaveBeenCalledWith('Failed to process Recurring Payment for trans-1', expect.any(Error))
```
irisfaraway, line 624: Would be good to make sure the same error you set up earlier is the one being called here. Otherwise this could get a false pass from throwing a different error.

packages/recurring-payments-job/src/__tests__/recurring-payments-processor.spec.js
```diff
@@ -588,6 +605,112 @@ describe('recurring-payments-processor', () => {
...
+      rpResultArgs: salesApi.processRPResult.mock.calls,
+      firstError: errorSpy.mock.calls[0]
+    }
+
+    expect(errorSummary).toEqual({
+      rpResultArgs: [
+        ['trans-1', paymentId1, date1],
+        ['trans-2', paymentId2, date2]
+      ],
+      firstError: ['Failed to process Recurring Payment for trans-1', expect.any(Error)]
```
irisfaraway, line 682: See comment above about testing for specific errors!

PR #2216, opened 2025-09-29
RP not updating CRM connectors lib

ScottDormand96:
https://eaflood.atlassian.net/browse/IWTF-4816

update the connectors lib for calling sales api from recurring pay job

### DIFF
--- a/packages/connectors-lib/src/sales-api-connector.js
+++ b/packages/connectors-lib/src/sales-api-connector.js
@@ -108,8 +108,8 @@ export const finaliseTransaction = async (id, payload) => exec2xxOrThrow(call(ne
  * @returns {Promise<*>}
  * @throws on a non-2xx response
  */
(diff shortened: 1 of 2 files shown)


packages/connectors-lib/src/sales-api-connector.js
```diff
@@ -27,7 +27,7 @@ export const call = async (url, method = 'get', payload = null) => {
     ok: response.ok,
     status: response.status,
     statusText: response.statusText,
-    body: response.status !== 204 ? await parseResponseBody(response) : undefined
+    body: response.status === 204 ? undefined : await parseResponseBody(response)
```
jaucourt: I'm not saying to change this, but simply so you're aware, you could also do:
```
  body: response.status === 204 && await parseResponseBody(response)
```
which would be slightly more succinct, but perhaps not quite as clear in intent

PR #2309, opened 2026-02-16
Cancellation journey page 3 functionality

gkoos:
https://eaflood.atlassian.net/browse/IWTF-4775


packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -170,7 +170,6 @@ export const cancelRecurringPayment = async (id, reason) => {
   const recurringPayment = await findById(RecurringPayment, id)
   if (recurringPayment) {
     const data = recurringPayment
-    data.cancelledDate = new Date().toISOString().split('T')[0]
     data.cancelledReason = await getGlobalOptionSetValue(RecurringPayment.definition.mappings.cancelledReason.ref, reason)
```
jaucourt: What Iris had said was "don't send the cancelledDate from the SalesAPI if the cancelledReason is 'User cancelled'." You've removed it for all cases here.

You'd also need to test both cases.

jaucourt (changes requested):

packages/sales-api-service/src/services/recurring-payments.service.js
gkoos: done

jaucourt (approved):

jaucourt (commented):
Please let Iris check this too

irisfaraway (approved):

PR #2332, opened 2026-03-23
GAFL: DEV - Handle Postcodes with >100 Results via Pagination (Offset) Before Filtering

gkoos:
https://eaflood.atlassian.net/browse/IWTF-5016

As a user entering my address for a dense postcode,
I want the system to fetch all addresses from the OS API (not just the first page),
so that my house name/number can be matched even when there are more than 100 addresses in the postcode.

### DIFF
--- a/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
+++ b/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
@@ -48,8 +48,8 @@ describe('address-lookup-service', () => {
                   BUILDING_NAME: buildingName,
                   THOROUGHFARE_NAME: thoroughfare,
                   DEPENDENT_LOCALITY: locality,
-                  POST_TOWN: town,
-              }
+                  POST_TOWN: town
+                }
               }
             ]
           })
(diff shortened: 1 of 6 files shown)


packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -517,6 +517,71 @@ describe('address-lookup-service', () => {
...
+      fetch.mockResolvedValueOnce({
+        json: () => ({
+          results: [
+            {
+              DPA: {
+                ADDRESS: '1 TEST STREET, BRISTOL, BS1 1AA',
+                POSTCODE: 'BS1 1AA',
+                BUILDING_NAME: '1 TEST STREET',
+                POST_TOWN: 'BRISTOL',
+                COUNTRY_CODE_DESCRIPTION: 'This record is within England'
```
irisfaraway: I thought we had gotten rid of COUNTRY_CODE_DESCRIPTION? There are a few mentions of it in here

irisfaraway (changes requested):
I know SonarQube caused a bit of pain by complaining about some functions containing too much complexity... and unfortunately I'm also going to be a bit of a pain by suggesting it gets broken down further. There is rather a lot going on here and breaking it up into smaller chunks would make it easier to follow everything that's going on (at least for my brain, anyway!) Even moving some of the one-liners into discrete, clearly-named functions would be beneficial, I think.

irisfaraway (approved):

PR #2358, opened 2026-04-08
Cancel agreement via Sales API

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5037

We want to notify GovPay when an agreement is cancelled so that we’re not relying on the CRM for this logic


packages/sales-api-service/src/schema/recurring-payments.schema.js
```diff
@@ -35,5 +35,5 @@ export const cancelRecurringPaymentRequestParamsSchema = Joi.object({
 })
 
 export const cancelRecurringPaymentRequestQuerySchema = Joi.object({
-  reason: Joi.string().required().valid('Payment Failure', 'User Cancelled')
+  reason: Joi.string().required().valid('Payment Failure', 'User Cancelled', 'Business Cancelled')
```
irisfaraway, line 38: Can we add a test to make sure this validates now?

packages/connectors-lib/src/__tests__/govuk-pay-api.spec.js
```diff
@@ -203,4 +203,30 @@ describe('govuk-pay-api-connector', () => {
...
+      })
+    })
+
+    it('logs and throws errors', async () => {
+      const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(jest.fn())
+      fetch.mockImplementation(() => {
+        throw new Error('cancel error')
+      })
+      await expect(govUkPayApi.cancelRecurringPaymentAgreement(123)).rejects.toEqual(Error('cancel error'))
+      expect(consoleErrorSpy).toHaveBeenCalled()
```
irisfaraway: Would prefer to confirm what error message the spy has been called with

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -170,21 +172,57 @@ export const cancelRecurringPayment = async (id, reason) => {
...
+    select: ['_defra_activepermission_value']
+  })
+  const permissionId = record._defra_activepermission_value
+  if (permissionId) {
+    return findById(Permission, permissionId)
+  }
+  return null
+}
+
+const cancelGovPayAgreement = async agreementId => {
```
irisfaraway: Suspect this is just me being a massive pedant, but it's GOV.UK Pay, not GovPay - I know we've been a bit inconsistent about this, but I'd prefer to use the full name, especially if the connector is called govUkPayApi. Same goes for various console logs, test names, etc

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -170,21 +172,57 @@ export const cancelRecurringPayment = async (id, reason) => {
...
-    await persist([updatedRecurringPayment])
+    const entitiesToPersist = [updatedRecurringPayment]
+
+    const linkedPermission = await getLinkedPermission(id)
+    if (linkedPermission) {
+      linkedPermission.isRecurringPayment = false
+      entitiesToPersist.push(linkedPermission)
+    }
+
+    await persist(entitiesToPersist)
```
irisfaraway: Just wondering what kind of logging/error handling we have if the cancellation fails at this stage, since would be awkward to have a situation where it's cancelled on GOV.UK Pay but not in the CRM. I _think_ it'll go just into the queue and try to save again later, but might be worth doing some testing around it.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
...
+    it('should not call cancelRecurringPaymentAgreement on GovPay when agreementId does not exist', async () => {
+      const recurringPayment = getMockRecurringPayment({ agreementId: undefined })
+      findById.mockReturnValueOnce(recurringPayment)
+
+      await cancelRecurringPayment('id', 'Payment Failure')
+
+      expect(govUkPayApi.cancelRecurringPaymentAgreement).not.toHaveBeenCalled()
+    })
+
+    it('should not throw when GovPay returns 404 (agreement already cancelled)', async () => {
```
irisfaraway: Does it really return a 404 if the agreement has already been cancelled? Seems odd that we wouldn't be able to distinguish between a cancelled agreement and the agreement ID just being wrong

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
         expect.objectContaining({
           ...recurringPayment,
           cancelledReason,
-          cancelledDate: null
-        })
+          cancelledDate: expect.stringMatching(/^\d{4}-\d{2}-\d{2}$/)
+        }),
+        mockPermission
```
irisfaraway: I don't think we need this test any more since there shouldn't be any difference in behaviour between these reasons now

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
...
+    it('should call cancelRecurringPaymentAgreement on GovPay when agreementId exists', async () => {
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+
+      expect(govUkPayApi.cancelRecurringPaymentAgreement).toHaveBeenCalledWith(recurringPayment.agreementId)
+    })
+
+    it('should not call cancelRecurringPaymentAgreement on GovPay when agreementId does not exist', async () => {
```
irisfaraway: This makes sense to me - but also makes me think that if we aren't doing the cancellation in GOV.UK Pay, we probably shouldn't be doing the rest of the cancellation stuff either. So maybe this should be a condition that just throws an error for all of `cancelRecurringPayment()` instead

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
...
+
+    it('should not throw when GovPay returns 400 (agreement in invalid state)', async () => {
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+      govUkPayApi.cancelRecurringPaymentAgreement.mockResolvedValueOnce({ ok: false, status: 400, statusText: 'Bad Request' })
+
+      await expect(cancelRecurringPayment('id', 'User Cancelled')).resolves.toBeDefined()
+    })
+
+    it('should still persist to CRM when GovPay returns 400', async () => {
```
irisfaraway: Do we want to do this? I'm concerned this could mean the CRM would show a record as cancelled when the agreement might still be active in GOV.UK Pay (although then I guess we would still not be requesting the payment when it would have been due, so maybe it's the same result). How does the CRM handle this situation currently?

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
...
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+      govUkPayApi.cancelRecurringPaymentAgreement.mockResolvedValueOnce({ ok: false, status: 400, statusText: 'Bad Request' })
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+      expect(persist).toHaveBeenCalled()
```
irisfaraway: Can we check what `persist()` is being called with?

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -970,11 +985,128 @@ describe('recurring payments service', () => {
...
+      await expect(cancelRecurringPayment('id', 'User Cancelled')).resolves.toBeDefined()
+    })
+
+    it('should still persist to CRM when GovPay returns 404', async () => {
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+      govUkPayApi.cancelRecurringPaymentAgreement.mockResolvedValueOnce({ ok: false, status: 404, statusText: 'Not Found' })
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+      expect(persist).toHaveBeenCalled()
```
irisfaraway: Can we check what `persist()` is being called with?

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
irisfaraway: Same goes for handling the 404 response above

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -170,21 +172,57 @@ export const cancelRecurringPayment = async (id, reason) => {
...
     throw new Error('Invalid id provided for recurring payment cancellation')
   }
 }
 
+const getLinkedPermission = async recurringPaymentId => {
+  const record = await dynamicsClient.retrieveRequest({
+    key: recurringPaymentId,
+    collection: RecurringPayment.definition.dynamicsCollection,
+    select: ['_defra_activepermission_value']
+  })
```
irisfaraway: Usually we define these queries in dynamics-lib, then call them from the applications as needed through the `dynamicsClient`. Have a look at `findLinkedRecurringPayment()` further down in this file for an example of the pattern in action.

irisfaraway (changes requested):

packages/sales-api-service/src/services/recurring-payments.service.js
lailien3: If it fails here after govuk pay cancellation succeeds, the error will come up as 500. Upon retry, there will be a 404 error which ive got handling for, so the persisit will run again successfully 🤞🏼 I will add some testing though

PR #2363, opened 2026-04-10
Address lookup is not populating premises field

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5100

This bug is not on production yet, but must be fixed before we can do a release.

When we use the address lookup to find an address, we can select an address by its house number or premises, but this field is then not being saved to the permission. This means that when we try to create a transaction in DynamoDb before taking payment, the permission fails validation and the journey cannot be continued.

We need to make sure the premises field is being set correctly, and test that the full journey works afterwards and the full address is being saved to the CRM. Unit tests should also be written to cover this.

### DIFF
--- a/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
+++ b/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
@@ -66,6 +66,68 @@ describe('address-lookup-service', () => {
         })
       }
     )
+
+    it('maps BUILDING_NUMBER to premises when BUILDING_NAME is absent', async () => {
+      fetch.mockResolvedValue({
+        json: () => ({
+          results: [
+            {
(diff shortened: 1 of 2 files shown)


packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -66,6 +66,68 @@ describe('address-lookup-service', () => {
...
+                BUILDING_NUMBER: '19',
+                THOROUGHFARE_NAME: 'CRABTREE AVENUE',
+                POST_TOWN: 'WEMBLEY'
+              }
+            }
+          ]
+        })
+      })
+      const results = await addressLookupService()
+      expect(results[0].premises).not.toBe('')
```
irisfaraway: Think this should specify the result we want, rather than just "not blank"

PR #2379, opened 2026-04-23
Update dynamics client config + token refresh

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5050

Currently, our config uses a callback in onTokenRefresh. In v2, callbacks will be removed, so we need to return the token instead.
Config fields like webApuUrl and webApiVersion need updating to the v2 format


packages/dynamics-lib/src/client/dynamics-client.js
```diff
@@ -21,14 +21,16 @@ export function config () {
 
   let accessToken = null
   return {
-    webApiUrl: process.env.DYNAMICS_API_PATH,
-    webApiVersion: process.env.DYNAMICS_API_VERSION,
+    serverUrl: new URL(process.env.DYNAMICS_API_PATH).origin,
```
irisfaraway: Rather than extract part of this URL, is it worth just updating the env vars to be the root domain and not include all the `/api/data/v9.1/` stuff? I think this is the only place it's used anyway.

irisfaraway (approved):

PR #2382, opened 2026-04-23
Update method call names

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5052
https://eaflood.atlassian.net/browse/IWTF-5053

This PR has merged these 2 tickets as they both cover updating the method call names, but one is for the mock helper. I did them together as they were quite entangled, and I needed them done in tandem to ensure I did it right.

### DIFF
--- a/packages/dynamics-lib/src/__mocks__/dynamics-web-api-mock-helper.js
+++ b/packages/dynamics-lib/src/__mocks__/dynamics-web-api-mock-helper.js
@@ -3,7 +3,7 @@ const { readFileSync } = jest.requireActual('fs')
 const Path = jest.requireActual('path')
 const optionSetDataPath = Path.join(Project.root, 'src', '__mocks__', 'option-set-data.json')
 
-export const configureDynamicsWebApiMock = (DynamicsWebApi = jest.genMockFromModule('dynamics-web-api')) => {
+export const configureDynamicsWebApiMock = (DynamicsWebApi) => {
   let expectedResponse = {}
   let nextResponses = {}
   let callError = {}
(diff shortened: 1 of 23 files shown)


packages/dynamics-lib/src/client/dynamics-client.js
```diff
@@ -21,14 +21,16 @@ export function config () {
 
   let accessToken = null
   return {
-    webApiUrl: process.env.DYNAMICS_API_PATH,
-    webApiVersion: process.env.DYNAMICS_API_VERSION,
+    serverUrl: process.env.DYNAMICS_API_PATH ? new URL(process.env.DYNAMICS_API_PATH).origin : undefined,
```
irisfaraway: Should this change be happening over in https://github.com/DEFRA/rod-licensing/pull/2379? We'd probably need some additional testing for this more complicated version, although I'm not sure we really need this extra fallback - if that env var isn't set then we probably want to be throwing an error here.

Same comment from that PR applies as well.

packages/dynamics-lib/src/__mocks__/dynamics-web-api-mock-helper.js
```diff
@@ -53,12 +53,14 @@ export const configureDynamicsWebApiMock = (DynamicsWebApi = jest.genMockFromMod
     }
   }
 
-  DynamicsWebApi.prototype.retrieveRequest = jest.fn(async () => responseCapableMethod('retrieveRequest'))
-  DynamicsWebApi.prototype.createRequest = jest.fn(async () => responseCapableMethod('createRequest'))
-  DynamicsWebApi.prototype.updateRequest = jest.fn(async () => responseCapableMethod('updateRequest'))
-  DynamicsWebApi.prototype.retrieveMultipleRequest = jest.fn(async () => responseCapableMethod('retrieveMultipleRequest'))
-  DynamicsWebApi.prototype.retrieveGlobalOptionSets = jest.fn(async () => responseCapableMethod('retrieveGlobalOptionSets'))
-  DynamicsWebApi.prototype.executeUnboundFunction = jest.fn(async functionName => {
+  DynamicsWebApi.prototype.retrieve = DynamicsWebApi.retrieve = jest.fn(async () => responseCapableMethod('retrieve'))
```
irisfaraway: Was curious about the switch to multiple assignments here - is this due to a breaking change in the upgrade?

irisfaraway (changes requested):

packages/dynamics-lib/src/client/dynamics-client.js
lailien3: it caused errors by this stage as i created each branch off the other - so i just fixed it here

packages/dynamics-lib/src/__mocks__/dynamics-web-api-mock-helper.js
lailien3: ah i forgot to remve them, i added them in while i was testing mock calls 😬

irisfaraway (approved):

PR #2397, opened 2026-05-12
SEMVER-PATCH : Release v1.70.1, RELEASE_TYPE=minor

lailien3:
Users reaching close to senior status are incorrectly being labelled as senior when renewing their licence.

### DIFF
--- a/packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
+++ b/packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
@@ -162,12 +162,12 @@ describe('preparePermissionDataForRenewal', () => {
     })
 
     it("doesn't add senior concession if the licensee turns 66 during the licence period but is 65 at the start", async () => {
-      const startDate = moment()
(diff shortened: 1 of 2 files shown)


packages/sales-api-service/src/services/renewals/__tests__/renewals.service.spec.js
```diff
@@ -161,6 +161,19 @@ describe('preparePermissionDataForRenewal', () => {
       expect(permission.concessions).toEqual([])
     })
 
+    it("doesn't add senior concession if the licensee turns 66 during the licence period but is 65 at the start", async () => {
+      const startDate = moment()
+      const birthDate = moment(startDate).subtract(65, 'years').subtract(6, 'months').format('YYYY-MM-DD')
```
jaucourt: don't rely on this sort of arithmetic - use actual dates. For example, a licence start date of '2026-05-12' and a angler birth date of '1960-09-01'

Also - do a few (it.each)

jaucourt (changes requested):

jaucourt (approved):

PR #2399, opened 2026-05-19
Cancel agreement via Sales API

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5037

We want to notify GovPay when an agreement is cancelled so that we’re not relying on the CRM for this logic


lailien3:
> Do we want to do this? I'm concerned this could mean the CRM would show a record as cancelled when the agreement might still be active in GOV.UK Pay (although then I guess we would still not be requesting the payment when it would have been due, so maybe it's the same result). How does the CRM handle this situation currently?



from what I get from the docs, a 400 specifically means "You can only cancel agreements in the active status", so by the time we get that response, the agreement is in a non-active state (created, cancelled, or inactive). there shouldn't be a case where the agreement is still active but we're marking it cancelled in CRM


My thinking was that if GovUK Pay already considers it non-active, persisting the cancellation to CRM just keeps things in sync and stops the recurring payments job from repeatedly trying to process it, but lmk if im thnking of this incorrectly

I've updated the test description tho to make that clearer


with the 404 error, in the docs too, 404 will mean that "No agreement matched the AGREEMENT_ID", so it's the ID not existing at all, not it being already cancelled. I also changed the test description to be clearer. For this, im thinking that becaus e we get the agreementId from CRM, so it was valid at one point, the 404 means it will be gone from GOVUK Pay's side. It feels safe to persist the cancellation to CRM, and no payments can be taken against it anyway. wdyt?

packages/connectors-lib/src/govuk-pay-api.js
```diff
@@ -113,3 +113,21 @@ export const getRecurringPaymentAgreementInformation = async agreementId => {
...
+ */
+export const cancelRecurringPaymentAgreement = async agreementId => {
+  try {
+    return fetch(`${process.env.GOV_PAY_RCP_API_URL}/${agreementId}/cancel`, {
+      headers: headers(true),
+      method: 'post',
+      timeout: process.env.GOV_PAY_REQUEST_TIMEOUT_MS || GOV_PAY_REQUEST_TIMEOUT_MS_DEFAULT
+    })
+  } catch (err) {
+    console.error(`Error cancelling recurring payment agreement in the GOV.UK API service - agreementId: ${agreementId}`, err)
```
jaucourt, line 130: If you're hoping to catch errors from the fetch with this, it won't work. Bonus points if you can tell me why! Super bonus points if you write a test for the behaviour... ;P 

Edit: I got AI to check the function and it didn't pick up the problem either!

packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
```diff
@@ -195,6 +179,11 @@ describe('cancelRecurringPaymentRequestQuerySchema', () => {
     expect(() => cancelRecurringPaymentRequestQuerySchema.validateAsync(sampleData)).not.toThrow()
   })
 
+  it('validates expected object if the reason is Business Cancelled', async () => {
+    const sampleData = { reason: 'Business Cancelled' }
+    expect(() => cancelRecurringPaymentRequestQuerySchema.validateAsync(sampleData)).not.toThrow()
+  })
```
jaucourt, line 175: Combining this with the Payment Failure test would be a good idea - and then you can add a case for User Cancelled, as that doesn't look to be tested for

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,241 @@ describe('recurring payments service', () => {
   })
 
   describe('cancelRecurringPayment', () => {
+    const mockPermission = new Permission()
+    mockPermission.isRecurringPayment = true
```
jaucourt: This is breaking the 'fresh test, fresh mocks' rule that we have

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,241 @@ describe('recurring payments service', () => {
...
-      await cancelRecurringPayment('id', 'Payment Failure')
+      await cancelRecurringPayment('id', 'User Cancelled')
 
       expect(persist).toHaveBeenCalledWith([
         expect.objectContaining({
           ...recurringPayment,
           cancelledReason,
           cancelledDate: expect.stringMatching(/^\d{4}-\d{2}-\d{2}$/)
-        })
+        }),
```
jaucourt, line 959: The cancelledDate isn't specific enough - this could be set to a static value that matched the regex. You need to mock out the date to be a couple of different values and test that it gets set to the current date / time. You're also only setting it to be a date, not a date / time.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,241 @@ describe('recurring payments service', () => {
...
+      findById.mockReturnValueOnce(recurringPayment)
 
-      const cancelledReason = { description: 'User Cancelled', id: 910400003, label: 'User Cancelled' }
+      await cancelRecurringPayment('id', 'User Cancelled')
 
-      const recurringPayment = { ...getMockRecurringPayment(), cancelledDate: null }
+      expect(govUkPayApi.cancelRecurringPaymentAgreement).toHaveBeenCalledWith(recurringPayment.agreementId)
+    })
+
+    it('should not call cancelRecurringPaymentAgreement on GovUKPay when agreementId does not exist', async () => {
```
jaucourt: This description could be a little more precise - it throws an error, so the description should say that we expect cancelRecurringPayment to throw an error when a fetched RCP doesn't have an agreement id

jaucourt (changes requested):
We need to chat about this - please give me a shout when you come to looking at review comments and we can pair up. You're not doing TDD and it's really confusing reading your tests and code because of that.

packages/connectors-lib/src/govuk-pay-api.js
lailien3: u need to await the fetch so the try/catch actually works!

packages/dynamics-lib/src/queries/__tests__/recurring-payments-queries.spec.js
```diff
@@ -62,4 +63,29 @@ describe('Recurring Payment Queries', () => {
...
+
+    it('includes recurring payment id filter', () => {
+      const request = getRetrieveRequest()
+      expect(request.filter).toContain(`defra_recurringpaymentid eq ${recurringPaymentId}`)
+    })
+
+    it('includes default statecode filter', () => {
+      const request = getRetrieveRequest()
+      expect(request.filter).toContain('statecode eq 0')
+    })
```
jaucourt, line 87: I can see what you've done with this test to break it down, and that's great. However, you could remove all instances of 'and' from the filter and no tests would fail, but it wouldn't work when used. I think either a snapshot test for the filter, with a couple of different values for recurring payment id, or a simple test against a string where you substitute in the recurringPaymentId, would be a bit more robust. Given that RecurringPayment.definition.defaultFilter isn't actually contained in the unit being tested, but we're asserting on this value, it should really be mocked out.

packages/dynamics-lib/src/queries/__tests__/recurring-payments-queries.spec.js
```diff
@@ -62,4 +63,29 @@ describe('Recurring Payment Queries', () => {
...
+
+    it('expands active permission relationship', () => {
+      const request = getRetrieveRequest()
+      expect(request.expand).toContainEqual({ property: 'defra_ActivePermission' })
+    })
```
jaucourt, line 94: Again, this is fine but you should be mocking out RecurringPayment otherwise you're testing more than the recurring payment query unit, you're testing the recurring payment entity too.

packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
```diff
@@ -168,50 +168,29 @@ describe('cancelRecurringPaymentRequestParamsSchema', () => {
...
-    const sampleData = { reason: 'Because I said so' }
-    expect(() => cancelRecurringPaymentRequestQuerySchema.validateAsync(sampleData).rejects.toThrow())
+  it('throws an error if reason is invalid', async () => {
+    const sampleData = { reason: 'Not A Valid Reason' }
+    await expect(cancelRecurringPaymentRequestQuerySchema.validateAsync(sampleData)).rejects.toThrow()
   })
 
   it('throws an error if an incorrect key is provided', async () => {
     const sampleData = { foo: 'Payment Failure' }
-    expect(() => cancelRecurringPaymentRequestQuerySchema.validateAsync(sampleData).rejects.toThrow())
```
jaucourt, line 215: Of course, you don't _have_ to use `validateAsync`. If you simply did `validate`, you get an object back with the validation details: https://joi.dev/api/18.x.x#anyvalidatevalue-options.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,238 @@ describe('recurring payments service', () => {
...
+      dynamicsClient.retrieveMultipleRequest.mockResolvedValue({
+        value: [{ defra_ActivePermission: { defra_permissionid: 'mock-permission-id' } }]
+      })
+      jest.spyOn(Permission, 'fromResponse').mockReturnValue(mockPermission)
+      retrieveGlobalOptionSets.mockReturnValue({ cached: jest.fn().mockResolvedValue({ definition: 'mock-def' }) })
+      findById.mockImplementation(entityType => {
+        if (entityType === Permission) return mockPermission
+        return null
+      })
+    })
```
jaucourt: This isn't a pattern we use. Please re-write this so that it's a helper function, into which you can pass a permission that you either create in the test, or is generated by another helper function. We can have a chat about this if what I'm saying isn't clear.

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -168,20 +171,57 @@ export const findNewestExistingRecurringPaymentInCrm = async agreementId => {
...
-    const data = recurringPayment
-    const isUserCancelled = reason === 'User Cancelled'
+  if (!recurringPayment) {
+    throw new Error('Invalid id provided for recurring payment cancellation')
+  }
+  if (!recurringPayment.agreementId) {
+    throw new Error('Cannot cancel a recurring payment without an agreement ID')
+  }
+
+  const data = recurringPayment
```
jaucourt: I don't understand why you're changing the name of the variable here? If you want it to be 'data', then give it that name to start with? Although I think calling it recurringPayment is more descriptive, and therefore better.

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -168,20 +171,57 @@ export const findNewestExistingRecurringPaymentInCrm = async agreementId => {
...
+
+  data.cancelledDate = new Date().toISOString()
+  data.cancelledReason = await getGlobalOptionSetValue(RecurringPayment.definition.mappings.cancelledReason.ref, reason)
+
+  await cancelGovUkPayAgreement(data.agreementId)
+
+  const updatedRecurringPayment = Object.assign(new RecurringPayment(), data)
```
jaucourt: Not really sure why you're doing this, either? Do the tests still pass if you don't?

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,238 @@ describe('recurring payments service', () => {
...
+      expect(persist).toHaveBeenCalledWith([expect.any(RecurringPayment), mockPermission])
+    })
+
+    it('passes recurring payment id to findPermissionByRecurringPaymentId', async () => {
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+
+      expect(findPermissionByRecurringPaymentId).toHaveBeenCalledWith('id')
```
jaucourt: please use a couple of different values for id, and don't be frightened of using a full guid

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,238 @@ describe('recurring payments service', () => {
...
     })
+
+    it('propagates persist error after successful GovUKPay cancellation', async () => {
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+      govUkPayApi.cancelRecurringPaymentAgreement.mockResolvedValueOnce({ ok: true, status: 204 })
+      persist.mockRejectedValueOnce(new Error('CRM unavailable'))
+
+      await expect(cancelRecurringPayment('id', 'User Cancelled')).rejects.toThrow('CRM unavailable')
+    })
```
jaucourt, line 1129: Much better & clearer tests - and they're comprehensive as far as I can see!

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
jaucourt: This comment is still relevant, because you're still using this static value in multiple tests...

jaucourt (changes requested):
Looking much better, but still a few things to address

packages/sales-api-service/src/schema/__tests__/recurring-payments.schema.spec.js
lailien3: I did try switching this block to `validate` and asserting specific validation detail types, but it ended up being brittle in CI (because of Joi error ordering/coercion) For `{ reason: 99 }`, I expected `string.base` but CI returned `any.only`, and for the unknown-key case I initially expected `object.unknown` but got `any.required` first.

I reverted this block back to `validateAsync(...).rejects.toThrow()`, which is also the pattern used elsewhere earlier in this same spec file.

packages/connectors-lib/src/govuk-pay-api.js
jaucourt: One tiny thing - you're still not awaiting the fetch. You've also not written a test for it...

packages/dynamics-lib/src/queries/__tests__/recurring-payments-queries.spec.js
```diff
@@ -62,4 +64,43 @@ describe('Recurring Payment Queries', () => {
...
+      const mockDefaultFilter = 'mock_default_filter'
+      const definitionSpy = jest.spyOn(RecurringPayment, 'definition', 'get').mockReturnValue(createMockDefinition(mockDefaultFilter))
+
+      try {
+        const request = findPermissionByRecurringPaymentId(recurringPaymentId).toRetrieveRequest()
+        expect(request.filter).toBe(
+          `_defra_activepermission_value ne null and defra_recurringpaymentid eq ${recurringPaymentId} and ${mockDefaultFilter}`
+        )
+      } finally {
+        definitionSpy.mockRestore()
```
jaucourt: Not 100% sure you need to do this - I think the mock might get torn down with the test.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +899,258 @@ describe('recurring payments service', () => {
...
-      retrieveGlobalOptionSets.mockReturnValueOnce({ cached: jest.fn().mockResolvedValue({ definition: 'mock-def' }) })
+    const cancelRecurringPaymentIds = [
+      '92207e2c-0aa2-4600-a9c4-a7eb2276e404',
+      'f794dd0d-b80a-4d51-b738-479c87b4765b',
+      'b437d543-430f-4eb1-830d-539148c8b440'
+    ]
+
+    const testState = {
+      linkedPermission: null
+    }
```
jaucourt: This, and the above, is sharing state between tests and is to be avoided. Each test should define its own mocks and its own state. Otherwise you risk ending up in a right old muddle, with tests passing that shouldn't, tests failing that shouldn't, and code that's difficult to read and understand. I think we should pair on this, so please put an hour in over the next day or two for us to go through it 😊

jaucourt (changes requested):

packages/connectors-lib/src/__tests__/govuk-pay-api.spec.js
```diff
@@ -202,5 +265,100 @@ describe('govuk-pay-api-connector', () => {
...
+      const error = new Error('cancel agreement failed')
+      const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(jest.fn())
+      fetch.mockRejectedValueOnce(error)
+
+      await govUkPayApi.cancelRecurringPaymentAgreement('abc-123').catch(() => {})
+      expect(consoleErrorSpy).toHaveBeenCalledWith(
+        'Error cancelling recurring payment agreement in the GOV.UK API service - agreementId: abc-123',
+        error
+      )
+    })
```
jaucourt, line 362: One small thing - you're always invoking `cancelRecurringPaymentAgreement` with `'abc-123'`. Change the containing describe block to have a few test cases with different agreement ids, so that it's tested with different values. Make sure that you change the description for the describe block to include the agreement id

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -168,20 +171,55 @@ export const findNewestExistingRecurringPaymentInCrm = async agreementId => {
...
   const recurringPayment = await findById(RecurringPayment, id)
-  if (recurringPayment) {
-    const data = recurringPayment
-    const isUserCancelled = reason === 'User Cancelled'
+  if (!recurringPayment) {
+    throw new Error('Invalid id provided for recurring payment cancellation')
+  }
+  if (!recurringPayment.agreementId) {
+    throw new Error('Cannot cancel a recurring payment without an agreement ID')
+  }
```
jaucourt, line 183: Both of these errors are great, but please can you add the recurring payment id to the error message? Then we'll have some chance of tracking it down if they're ever thrown.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +898,257 @@ describe('recurring payments service', () => {
   })
 
   describe('cancelRecurringPayment', () => {
-    it('should call findById with RecurringPayment and the provided id', async () => {
-      retrieveGlobalOptionSets.mockReturnValueOnce({ cached: jest.fn().mockResolvedValue({ definition: 'mock-def' }) })
+    const cancelRecurringPaymentIds = [
+      '92207e2c-0aa2-4600-a9c4-a7eb2276e404',
+      'f794dd0d-b80a-4d51-b738-479c87b4765b',
+      'b437d543-430f-4eb1-830d-539148c8b440'
+    ]
```
jaucourt: Don't do this, please - just put three different guids in the each of the relevant test. And make them different between tests, because there's no need to share guids in this way.

packages/sales-api-service/src/services/recurring-payments.service.js
```diff
@@ -168,20 +171,55 @@ export const findNewestExistingRecurringPaymentInCrm = async agreementId => {
...
+  await cancelGovUkPayAgreement(recurringPayment.agreementId)
+
+  const updatedRecurringPayment = Object.assign(new RecurringPayment(), recurringPayment)
+  const entitiesToPersist = [updatedRecurringPayment]
+
+  const linkedPermission = await getLinkedPermission(id)
+  if (linkedPermission) {
+    linkedPermission.isRecurringPayment = false
+    entitiesToPersist.push(linkedPermission)
+  }
```
jaucourt: This could be done more efficiently by retrieving the recurring payment with the active permission expanded (appending `$expand=defra_ActivePermission` to the oData query), then only one round trip is needed.

I'm also not sure under what circumstances there _wouldn't_ be an active permission? I think perhaps an error should be thrown in this circumstance.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +898,257 @@ describe('recurring payments service', () => {
...
+    it('returns the updated recurring payment', async () => {
+      arrangeCancelRecurringPayment()
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
 
-      const cancelledReason = { description: 'User Cancelled', id: 910400003, label: 'User Cancelled' }
+      const result = await cancelRecurringPayment('id', 'User Cancelled')
 
-      const recurringPayment = { ...getMockRecurringPayment(), cancelledDate: null }
+      expect(result).toEqual(expect.any(RecurringPayment))
```
jaucourt: I don't think this is specific enough. The `cancelRecurringPayment` could do `return new RecurringPayment()` and this test would still pass. You need to verify what's returned matches what you'd expect - i.e. the mock recurring payment you've created above.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +898,257 @@ describe('recurring payments service', () => {
...
+
+    it('builds linked permission from expanded response data', async () => {
+      const permissionData = { defra_permissionid: 'specific-permission-id' }
+      arrangeCancelRecurringPayment({ retrieveMultipleResponse: { value: [{ defra_ActivePermission: permissionData }] } })
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+
+      expect(Permission.fromResponse).toHaveBeenCalledWith(permissionData, expect.anything())
```
jaucourt: It's good you're testing that permissionData is passed to Permission.fromResponse, but you're not testing that the value from `retrieveGlobalOptionSets().cached()` is passed to it as the second argument.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +898,257 @@ describe('recurring payments service', () => {
...
+    })
+
+    it('persists recurring payment when no linked permission is found', async () => {
+      arrangeCancelRecurringPayment({ retrieveMultipleResponse: { value: [] } })
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+
+      await cancelRecurringPayment('id', 'User Cancelled')
+
+      expect(persist).toHaveBeenCalledWith([expect.any(RecurringPayment)])
```
jaucourt: Not specific enough - you could do `persist([new RecurringPayment()])` and the test would still pass. Same goes for the next test.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
```diff
@@ -895,93 +898,257 @@ describe('recurring payments service', () => {
...
     })
+
+    it('propagates persist error after successful GovUKPay cancellation', async () => {
+      arrangeCancelRecurringPayment()
+      const recurringPayment = getMockRecurringPayment()
+      findById.mockReturnValueOnce(recurringPayment)
+      persist.mockRejectedValueOnce(new Error('CRM unavailable'))
+
+      await expect(cancelRecurringPayment('id', 'User Cancelled')).rejects.toThrow('CRM unavailable')
+    })
```
jaucourt, line 1129: As well as my other comments, you're missing a few test cases:
1) If the object returned by findById doesn't have an agreementId property, an error is thrown
2) When cancelGovUkPayAgreement is called, if the response from govUkPayApi is ok, writes out a message and the agreement id via the debug statement
3) When cancelGovUkPayAgreement is called, if the response from govUkPayApi  has a status of not found, it writes out a message and the agreement id via the debug statement
4) When cancelGovUkPayAgreement is called, if the response from govUkPayApi  has a status of bad request, it writes out a message and the agreement id via the debug statement

You might be able to delete one or two tests if you amend the data retrieval so that the expanded active permission is retrieved. Although an error probably should be thrown if there is no active permission.

jaucourt (changes requested):
Sorry, still a few things to address.

packages/sales-api-service/src/services/__tests__/recurring-payments.service.spec.js
lailien3: that test is now gone cus we changed the refactored and changed the request to dynamics into 1 call - i believe ur concern is now covered in predefined-query.spec.js

PR #2402, opened 2026-05-20
Corrected link colours in RCP journey

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5123

Links in various pages of the cancellation journey are not in the correct colours per the Design System. This is in both the English and Welsh journeys. The bright blue below is not correct.


irisfaraway (approved):
LGTM! But hold off merging until release is sorted

PR #2410, opened 2026-05-28
Refactored renewal service doesn't handle junior concessions properly

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5145
https://eaflood.atlassian.net/browse/IWTF-5144

We recently refactored easy renewals to use the renewals service that we built for recurring payments. However because recurring payments aren't available for juniors, the service did not handle the transition from junior to adult licences properly. This PR fixes that bug so a user who has turned 17 since last purchasing a licence will be offered an adult licence when going through easy renewals.

It also fixes an issue where juniors with a disabled concession were being issued disabled adult licences.


packages/sales-api-service/src/services/__tests__/concession.service.spec.js
```diff
@@ -199,21 +199,29 @@ describe('preparePermissionDataForRenewal', () => {
...
         const permission = getMockPermission()
-        const disabledConcession = getDisabledNiConcession()
+        const disabledConcession = getDisabledBlueBadgeConcession()
         permission.concessions = [disabledConcession]
         await concessionService.addSenior(permission)
         expect(permission.concessions).toEqual([disabledConcession, getSeniorConcession()])
       })
 
-      it('add senior to NI', async () => {
+      it('adds a junior concession and keeps existing NI concession', async () => {
```
jaucourt: This name doesn't match what's happening - it's adding a senior concession to a disabled concession and then checking they're both applied to the permission after the `addSenior` call. There's nothing to do with juniors in the test.

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
```diff
@@ -199,21 +199,29 @@ describe('preparePermissionDataForRenewal', () => {
...
         await concessionService.addSenior(permission)
         expect(permission.concessions).toEqual([disabledConcession, getSeniorConcession()])
       })
+
+      it('removes any existing junior concession', async () => {
+        const permission = getMockPermission()
+        const juniorConcession = getJuniorConcession()
+        permission.concessions = [juniorConcession]
+        await concessionService.addSenior(permission)
+        expect(permission.concessions).toEqual([getSeniorConcession()])
```
jaucourt, line 223: This doesn't read quite right to me - I feel like you should be explicitly testing that the junior concession  is absent. Quite happy to be disagreed with, though, I know you're effectively doing this by ensuring the senior concession is the only concession applied.

packages/sales-api-service/src/services/renewals/renewals.service.js
```diff
@@ -101,6 +101,10 @@ const prepareConcessionsData = async (existingPermission, dateData) => {
 
   if (isSenior(ageAtLicenceStartDate)) {
     await addSenior(existingPermission)
+  } else if (isJunior(ageAtLicenceStartDate)) {
+    await addJunior(existingPermission)
+  } else {
+    await removeJunior(existingPermission)
```
jaucourt, line 107: Do we need a `removeSenior` here as well?

jaucourt (changes requested):

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
irisfaraway: Oops, yeah, that was a copy-paste error!

packages/sales-api-service/src/services/renewals/renewals.service.js
irisfaraway: Yeah, that probably wouldn't hurt. I'll add it.

packages/sales-api-service/src/services/__tests__/concession.service.spec.js
irisfaraway: I thought this was preferable since it also tests that the senior concession is successfully added, even if a junior one existed before.

jaucourt (approved):

PR #2414, opened 2026-06-08
License type on RCP details translation

Lawrence-Sithole-Def:
https://eaflood.atlassian.net/browse/IWTF-5125


Lawrence-Sithole-Def:
image

The test subject is treated as already “Senior” at licence start, but in the test they are still 65.

packages/gafl-webapp-service/src/pages/recurring-payments/cancel/details/__tests__/route.spec.js
```diff
@@ -125,45 +125,40 @@ describe('route', () => {
...
-        { key: { text: mssgs.rp_cancel_details_last_purchased }, value: { text: sampleData.permission.referenceNumber } },
-        { key: { text: mssgs.rp_cancel_details_licence_valid_until }, value: { text: sampleFormattedDate } }
-      ])
+      const licenceTypeRow = result.summaryTable.find(row => row.key.text === mssgs.rp_cancel_details_licence_type)
+      expect(licenceTypeRow.value.text).toContain('Trout and coarse (up to 2 rods)')
+      if (expectOver66) {
+        expect(licenceTypeRow.value.text).toContain('(over_66)')
+      } else {
+        expect(licenceTypeRow.value.text).not.toContain('(over_66)')
+      }
```
jaucourt: I'm a bit lost here. You've removed a test that was checking all rows of the summary table with one that only checks one row. So I think the other rows aren't tested now - which I don't think is right. Have I got this wrong?

packages/gafl-webapp-service/src/processors/recurring-payments-write-cache.js
```diff
@@ -11,7 +25,12 @@ export const setupCancelRecurringPaymentCacheFromAuthResult = async (request, au
         lastName: licensee.lastName,
         preferredMethodOfConfirmation: licensee.preferredMethodOfConfirmation?.label
       },
-      permit: { description: permit.description }
+      permit: {
+        description: permit.description,
+        permitSubtype: permit.permitSubtype,
+        numberOfRods: permit.numberOfRods
+      },
+      concessions: concessionsFromPermitDescription(permit.description)
```
jaucourt: Rather than extracting concessions from the permit description, can you not retrieve them from the concessionProofs linked to the existing permission?

PR #2436, opened 2026-07-06
Reintroduce address lookup refactor with fixes

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5133

This PR reverts commit a96a1fa15e334dc25d1d4d882ec13428f71d53bc. It will be used as the main feature branch for additional fixes needed to make the address lookup changes work as required.


docker/env/gafl_webapp_telesales.env.example
```diff
@@ -10,8 +10,8 @@ REDIS_HOST=host.docker.internal
 REDIS_PORT=6379
 
 # Address lookup
-ADDRESS_LOOKUP_URL=http://host.docker.internal:9002/address-service/v1/addresses/postcode-and-premises
-ADDRESS_LOOKUP_KEY=IWTFdev
+ADDRESS_LOOKUP_URL=https://api.os.uk/search/places/v1/postcode
+ADDRESS_LOOKUP_KEY=
```
jaucourt: Is this still required? You look to have deleted it from the websales env file.

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -1,50 +1,227 @@
...
-    premises: premises,
-    key: process.env.ADDRESS_LOOKUP_KEY,
-    lang: ADDRESS_LOOKUP_SERVICE.lang,
-    dataset: ADDRESS_LOOKUP_SERVICE.dataset
+    key: process.env.ADDRESS_LOOKUP_KEY
   })
 
+  if (offset > 0) {
+    params.append('offset', offset)
+  }
```
jaucourt, line 22: Tiny point, but would it be worth throwing an error if offset isn't greater than zero?

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -1,50 +1,227 @@
...
+ */
+const fetchAdditionalPages = async (postcode, totalresults, maxresults, cap) => {
+  const effectiveTotal = Math.min(totalresults, cap)
+
+  // Calculate offsets for additional pages (first page already fetched at offset 0)
+  // Example: if effectiveTotal=250 and maxresults=100, generates [100, 200] to fetch pages 2 and 3
+  const offsets = Array.from(
+    { length: Math.ceil((effectiveTotal - maxresults) / maxresults) },
+    (_, i) => maxresults + i * maxresults
+  ).filter(offset => offset < effectiveTotal)
```
jaucourt: Using Array.from like this is a bit unusual, because the spec specifies that the first element should be an Array or Array-like iterable. Whilst a simple object with a length property works, it'd be a bit more robust to do `new Array(Math.ceil((effectiveTotal - maxresults) / maxresults))`.

Also, whilst concise, it's also not the most readable bit of code I've ever seen - I had to transpose it into CodePen to understand what it was doing. But perhaps I'm just a bit thick!

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -1,50 +1,227 @@
...
+  const pageResults = await Promise.allSettled(offsets.map(offset => fetchPage(buildUrl(postcode, offset))))
+
+  const additionalResults = pageResults.filter(r => r.status === 'fulfilled' && r.value.results).flatMap(r => r.value.results)
+
+  const failedPages = pageResults
+    .map((result, idx) => ({ result, offset: offsets[idx] }))
+    .filter(({ result }) => result.status === 'rejected')
+    .map(({ result, offset }) => ({
+      offset,
+      error: result.reason?.message || 'Unknown error'
```
jaucourt, line 208: This is another function that could probably be a little clearer too...

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -1,50 +1,227 @@
...
+
+  // Fetch additional pages if needed
+  const cap = getMaximumResultsCap()
+  const {
+    results: additionalResults,
+    failedPages,
+    pagesFetched: additionalPagesFetched
+  } = needsPagination
+    ? await fetchAdditionalPages(postcode, totalresults, maxresults, cap)
+    : { results: [], failedPages: [], pagesFetched: 0 }
```
jaucourt: The combination of an object destructure as well as a ternary makes this a bit too complex to tease the meaning from. Please could you simplify it a bit?

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -1,50 +1,227 @@
...
+ * @param {object} firstPage - First page response
+ * @param {Array} additionalResults - Results from additional pages
+ * @param {Array} failedPages - Failed page information
+ * @param {number} additionalPagesFetched - Count of additional pages fetched
+ * @param {string} postcode - Postcode being searched
+ * @param {number} cap - Maximum results cap
+ * @param {number} startTime - Start timestamp for telemetry
+ * @returns {Array} Aggregated results
+ */
+const processResults = (firstPage, additionalResults, failedPages, additionalPagesFetched, postcode, cap, startTime) => {
```
jaucourt: It's an internal function that's only used once, but having so many parameters for a function isn't ideal. Perhaps you could use a config object?

jaucourt (changes requested):

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: I think it's a legit use of a sequence generator pattern: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#sequence_generator_range I have split it up and done a bit more commenting and renaming to make it a bit clearer though

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: We call it with 0 to get the first page. I don't think there's a situation where it would ever be less than 0

docker/env/gafl_webapp_telesales.env.example
irisfaraway: Fixed!

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: Refactored

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: Refactored

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: Have added some comments to this one

PR #2437, opened 2026-07-06
Build premises field based on OS Places results

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5134

We need to build a premises field for each address we get back from OS Places, matching the logic followed by the addressbase facade. This will be used for filtering and for eventually saving the chosen address.

### DIFF
--- a/packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
+++ b/packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
@@ -37,6 +37,35 @@ const fetchPage = async url => {
   return response.json()
 }
 
+const buildPremises = result => {
+  if (shouldBeOrganisationOnly(result)) {
+    return result.ORGANISATION_NAME
+  }
+
+  const premisesFields = []
+  if (isPoBoxAddress(result)) {
+    premisesFields.push('PO BOX ' + result.PO_BOX_NUMBER)
+  } else {
+    premisesFields.push(result.PO_BOX_NUMBER)
+  }
+  premisesFields.push(result.SUB_BUILDING_NAME, result.BUILDING_NAME, result.BUILDING_NUMBER)
+
+  return premisesFields.filter(Boolean).join(', ')
+}
+
+const shouldBeOrganisationOnly = result => {
+  const fieldsToCheck = [
+    result.CLASSIFICATION_CODE,
+    result.PO_BOX_NUMBER,
+    result.SUB_BUILDING_NAME,
+    result.BUILDING_NAME,
+    result.BUILDING_NUMBER
+  ]
+  return fieldsToCheck.filter(Boolean).length === 0
+}
+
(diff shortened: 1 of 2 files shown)


irisfaraway:
Chose to just log this out for testing purposes as subsequent tickets will actually work it into the code.

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -37,6 +37,35 @@ const fetchPage = async url => {
...
+
+const shouldBeOrganisationOnly = result => {
+  const fieldsToCheck = [
+    result.CLASSIFICATION_CODE,
+    result.PO_BOX_NUMBER,
+    result.SUB_BUILDING_NAME,
+    result.BUILDING_NAME,
+    result.BUILDING_NUMBER
+  ]
+  return fieldsToCheck.filter(Boolean).length === 0
```
jaucourt: I think this could be much clearer. I've worked out that it's trimming out empty fields (although it'll leave ones that only consist of spaces), but it should be clear from just reading it.

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -37,6 +37,35 @@ const fetchPage = async url => {
...
+
+  const premisesFields = []
+  if (isPoBoxAddress(result)) {
+    premisesFields.push('PO BOX ' + result.PO_BOX_NUMBER)
+  } else {
+    premisesFields.push(result.PO_BOX_NUMBER)
+  }
+  premisesFields.push(result.SUB_BUILDING_NAME, result.BUILDING_NAME, result.BUILDING_NUMBER)
+
+  return premisesFields.filter(Boolean).join(', ')
```
jaucourt: There are more expressive ways of finding only fields that have values

jaucourt (changes requested):

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
irisfaraway: I actually copied `filter(Boolean)` from elsewhere in our codebase, as we use it in a few places - it basically just checks if the value evaluates as true, which eliminates `undefined`s, `null`s and empty strings. I quite like it.

But you are right that it would not eliminate a string that only contains blank spaces. It also wouldn't handle an integer of 0 (which is occasionally used as a house name or number, although from what I've seen these are always returned as strings anyway). I've updated the filtering to be more explicit and cover more edge cases.

jaucourt: You're right - I see Gabor introduced one of the others, and Michael did the final one a few years ago. The more I think about it the more I dislike it, so I'm very glad you're changed it. Might go change the others now...

jaucourt (approved):

PR #2441, opened 2026-07-16
Recurring payments shouldn't be offered to minors (under 18) 

lailien3:
https://eaflood.atlassian.net/browse/IWTF-5168

Junior licences are currently not offers an rcp agreement, however their max age is 17. Only if the purchaser is 18 years old at the time of purchase should they be offered a recurring payment.

### DIFF
--- a/packages/business-rules-lib/src/util/ages.js
+++ b/packages/business-rules-lib/src/util/ages.js
@@ -2,6 +2,8 @@
 export const MINOR_MAX_AGE = 12
 /** The maximum age at which an angler is entitled to a junior concession */
 export const JUNIOR_MAX_AGE = 16
+/** The minimum age at which a recurring card payment can we offered */
+export const ADULT_MIN_AGE = 18
 /** The minimum age at which an angler becomes entitled to a senior concession */
 export const SENIOR_MIN_AGE = 66
 
@@ -18,6 +20,13 @@ export const isMinor = age => age <= MINOR_MAX_AGE
  * @returns {boolean} true if the given age should be classified as a junior
  */
 export const isJunior = age => age > MINOR_MAX_AGE && age <= JUNIOR_MAX_AGE
+
+/**
+ * Determine if the provided age is classified as an adult
+ * @param {number} age The age to be tested
+ * @returns {boolean} true if the given age should be classified as an adult
+ */
+export const isAdult = age => age >= ADULT_MIN_AGE
 
 /**
  * Determine if the provided age is classified as a senior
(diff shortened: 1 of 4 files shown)


packages/business-rules-lib/src/util/ages.js
```diff
@@ -2,6 +2,8 @@
 export const MINOR_MAX_AGE = 12
 /** The maximum age at which an angler is entitled to a junior concession */
 export const JUNIOR_MAX_AGE = 16
+/** The minimum age at which a recurring card payment can we offered */
```
irisfaraway:
```suggestion
/** The minimum age at which a recurring card payment can be offered */
```

Typo :)

irisfaraway (commented):
Implementation looks fine to me. My only question is around the naming - since we usually talk about the license 17 year old would be issued as an "adult licence", should we maybe call it `LEGAL_ADULT_MIN_AGE` or something instead, just to keep it distinct from the licence type? Happy to discuss though!

packages/gafl-webapp-service/src/processors/recurring-pay-helper.js
```diff
@@ -16,7 +16,7 @@ export const validForRecurringPayment = permission => {
   return (
     permission.licenceLength === '12M' &&
     permission.isLicenceForYou &&
-    licenseeAge > JUNIOR_MAX_AGE &&
+    licenseeAge >= ADULT_MIN_AGE &&
```
jaucourt: I don't like the naming convention you've used here. It is logical, but my worry is that this might end up getting used outside of this scope, and we could end up with some anglers being in a limbo of not being either juniors or adults. Please could you change the name to `RECURRING_PAYMENT_MIN_AGE` (or something like that)? Then there's less danger of confusion.

jaucourt (changes requested):

packages/business-rules-lib/src/util/ages.js
```diff
@@ -19,6 +21,13 @@ export const isMinor = age => age <= MINOR_MAX_AGE
  */
 export const isJunior = age => age > MINOR_MAX_AGE && age <= JUNIOR_MAX_AGE
 
+/**
+ * Determine if the provided age is classified as an adult
+ * @param {number} age The age to be tested
+ * @returns {boolean} true if the given age should be classified as an adult
+ */
+export const isAdult = age => age >= RECURRING_PAYMENT_MIN_AGE
```
irisfaraway: Sorry, naming again - I think this needs to be more explicit that it is checking the minimum payment recurring age, rather than "isAdult". The mentions of "classified as an adult" also sound like this could be about adult licences, which includes 17 year olds.

`isOldEnoughForRecurringPayment`, maybe?

lailien3: oh yes sorry i think i missed updating that one!

PR #2443, opened 2026-07-21
Filter OS Places results based on premises field

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5135
https://eaflood.atlassian.net/browse/IWTF-5136

The JS implementation of the address lookup currently filters based on the ADDRESS field that OS Places returns. Now that we have built our own custom premises field, we should use that to filter the search results instead.

Search results should then be ordered with the closest matches at the top (mimicking the logic used by the old Java implementation).

If the user selects the address, we should also save our newly-built custom premises as the premises.

### DIFF
--- a/docker/env/gafl_webapp.secrets.env.example
+++ b/docker/env/gafl_webapp.secrets.env.example
@@ -6,3 +6,4 @@ GOV_PAY_APIKEY=<insert here>
 GOV_PAY_RECURRING_APIKEY=<insert here>
 SESSION_COOKIE_PASSWORD=<insert here>
 ANALYTICS_PROPERTY_API=<insert here>
+ADDRESS_LOOKUP_KEY=<insert here>
--- a/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
+++ b/packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
@@ -1,12 +1,24 @@
 import addressLookupService from '../address-lookup-service.js'
 import fetch from 'node-fetch'
-import db from 'debug'
 
 jest.mock('node-fetch')
-jest.mock('debug', () => jest.fn(() => jest.fn()))
-const { value: debug } = db.mock.results[db.mock.calls.findIndex(c => c[0] === 'webapp:address-lookup-service')]
 
 describe('address-lookup-service', () => {
+  const createMockAddress = ({ buildingName, buildingNumber, classificationCode, organisationName, poBoxNumber, subBuildingName }) => ({
+    DPA: {
+      ADDRESS: 'FISHCORP, FISH BOULEVARD, FISHBOROUGH, FI1 5SH',
+      BUILDING_NUMBER: buildingNumber,
+      BUILDING_NAME: buildingName,
+      CLASSIFICATION_CODE: classificationCode,
+      ORGANISATION_NAME: organisationName,
+      PO_BOX_NUMBER: poBoxNumber,
+      POST_TOWN: 'BRISTOL',
+      POSTCODE: 'BS1 1AA',
+      SUB_BUILDING_NAME: subBuildingName,
+      THOROUGHFARE_NAME: 'TEST STREET'
+    }
+  })
+
   beforeAll(() => {
     process.env.ADDRESS_LOOKUP_KEY = 'ADDRESS_LOOKUP_KEY'
     process.env.ADDRESS_LOOKUP_URL = 'https://address.lookup.url'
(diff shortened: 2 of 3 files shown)


docker/env/gafl_webapp.secrets.env.example
```diff
@@ -6,3 +6,4 @@ GOV_PAY_APIKEY=<insert here>
 GOV_PAY_RECURRING_APIKEY=<insert here>
 SESSION_COOKIE_PASSWORD=<insert here>
 ANALYTICS_PROPERTY_API=<insert here>
+ADDRESS_LOOKUP_KEY=<insert here>
```
jaucourt, line 9: Please could you add an entry in the README for this as well?

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -71,574 +96,376 @@ describe('address-lookup-service', () => {
...
 
-    it('normalizes multiple spaces in both input and API results', async () => {
+    it.each`
+      snakeCaseField           | camelCaseField
+      ${'BUILDING_NAME'}       | ${'buildingName'}
+      ${'CLASSIFICATION_CODE'} | ${'classificationCode'}
+      ${'BUILDING_NUMBER'}     | ${'buildingNumber'}
+      ${'PO_BOX_NUMBER'}       | ${'poBoxNumber'}
+      ${'SUB_BUILDING_NAME'}   | ${'subBuildingName'}
+    `('does not match to ORGANISATION_NAME if $snakeCaseField is present', async ({ _snakeCaseField, camelCaseField }) => {
```
jaucourt: You don't need to put an underscore in front of _snakeCaseField - if fact, I think if you were to log it out, you'd see it's `undefined`. You can simply omit it as you're doing object destructuring.

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -71,574 +96,376 @@ describe('address-lookup-service', () => {
...
+        const addressResults = []
+        for (const address of disorderedAddresses) {
+          addressResults.push(createMockAddress({ buildingName: address }))
         }
-      ])
-    })
+        fetch.mockResolvedValueOnce({ json: () => ({ results: addressResults }) })
+
+        const results = await addressLookupService(exactMatch, 'FI1 5SH')
+        const orderedAddresses = [exactMatch, partialMatch, anotherPartialMatch, onlyMatchesABit]
```
jaucourt: Tiny point, but I think this belongs in the 'arrange' section, probably the line after the disorderedAddresses declaration.

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -71,574 +96,376 @@ describe('address-lookup-service', () => {
...
+
+        const results = await addressLookupService(term, 'FI1 5SH')
+        const orderedAddresses = [
+          matchesLettersAndDigits,
+          alsoMatchesLettersAndDigits,
+          matchesDigitsOnly,
+          matchesLettersOnly,
+          alsoMatchesLettersOnly,
+          alsoMatchesDigitsOnly
+        ]
```
jaucourt: Ah, same here - and for the following tests.

packages/gafl-webapp-service/src/services/address-lookup/address-lookup-service.js
```diff
@@ -71,29 +72,61 @@ const removeTrailingWhitespace = value => (typeof value === 'string' ? value.tri
...
+    .replaceAll(/\(|\)/g, '')
+    .split(/\s+|-|\.|,/)
+  const termsExcludingThe = terms.filter(term => term !== 'the')
+  return removeMissingOrBlankFields(termsExcludingThe)
+}
+
+const checkQualityOfMatch = (userProvidedSearchTerm, resultTerm) => {
+  if (userProvidedSearchTerm === resultTerm) {
+    // Matches that contain both letters and digits score higher
+    if (/[a-z]+/.test(userProvidedSearchTerm) && /\d+/.test(userProvidedSearchTerm)) {
```
jaucourt: You can accomplish this in a single regex by using positive lookaheads: `/^(?=.*[a-zA-Z])(?=.*\d)/` will check that both alphabetical characters and numerals exist in the string (in any order, so it'll match '123 Bramble Drive', 'Flat 7' or even "abc123abc').

jaucourt (changes requested):

docker/env/gafl_webapp.secrets.env.example
irisfaraway: It's already in there from Gabor's original code - it was just missing from the secrets.env.example

PR #2448, opened 2026-07-31
Save street fields from OS Places data

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5137

Once the user selects an address, we should save the street field based on the DEPENDENT_THOROUGHFARE_NAME and the THOROUGHFARE_NAME.


packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -7,6 +7,31 @@ jest.mock('debug', () => jest.fn(() => jest.fn()))
...
 describe('address-lookup-service', () => {
+  const createMockAddress = ({
+    buildingName,
+    buildingNumber,
+    classificationCode,
+    dependentThoroughfareName,
+    organisationName,
+    poBoxNumber,
+    subBuildingName,
+    thoroughfareName
```
jaucourt, line 15: You can omit the fields that're never used

jaucourt (changes requested):

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
irisfaraway: They're used for the premises field, so I figured I'd leave them as they'll just get rebased back in almost immediately

PR #2449, opened 2026-07-31
Save locality fields from OS Places

irisfaraway:
https://eaflood.atlassian.net/browse/IWTF-5138

Once the user selects an address, we should save the locality based on the DOUBLE_DEPENDENT_LOCALITY and DEPENDENT_LOCALITY.


packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
```diff
@@ -7,6 +7,35 @@ jest.mock('debug', () => jest.fn(() => jest.fn()))
...
+      CLASSIFICATION_CODE: classificationCode,
+      DEPENDENT_LOCALITY: dependentLocality,
+      DEPENDENT_THOROUGHFARE_NAME: dependentThoroughfareName,
+      DOUBLE_DEPENDENT_LOCALITY: doubleDependentLocality,
+      ORGANISATION_NAME: organisationName,
+      PO_BOX_NUMBER: poBoxNumber,
+      POST_TOWN: 'BRISTOL',
+      POSTCODE: 'BS1 1AA',
+      SUB_BUILDING_NAME: subBuildingName,
+      THOROUGHFARE_NAME: thoroughfareName
```
jaucourt, line 32: Most of these fields are never used so should be omitted, I think. You're not providing values so they'll be `undefined`, if the key doesn't exist it'll still be `undefined`

jaucourt (changes requested):

packages/gafl-webapp-service/src/services/address-lookup/__tests__/address-lookup-service.spec.js
irisfaraway: They're used for the premises and the street field, so I figured I'd leave them as they'll just get rebased back in almost immediately

